PR #11, opened 2022-05-24
Fix balance not updating on refresh

msalcala11:
Fixes the issue on the home screen and also on the key overview screen.

### DIFF
--- a/src/store/wallet/effects/status/status.ts
+++ b/src/store/wallet/effects/status/status.ts
@@ -293,13 +293,7 @@ export const startUpdateAllWalletStatusForKey =
                   return bStatus.walletId === wallet.id;
                 }) || {};
 
-              if (
-                !force ||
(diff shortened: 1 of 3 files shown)


src/store/wallet/effects/status/status.ts
```diff
@@ -293,13 +293,7 @@ export const startUpdateAllWalletStatusForKey =
                   return bStatus.walletId === wallet.id;
                 }) || {};
 
-              if (
-                !force ||
-                !status ||
-                !success ||
-                // skip formatting amounts if nothing has changed
-                status.balance.totalAmount === cachedBalance.sat
```
JohnathanWhite, line 301: we need to preserve this condition - it prevents unneeded status building

JohnathanWhite:
resolved in migration PR

PR #12, opened 2022-05-25
[FEAT] handle qrcode import and join

gabrielbazan7:


src/navigation/wallet/screens/JoinMultisig.tsx
```diff
@@ -153,28 +123,64 @@ const JoinMultisig = () => {
...
+                        params: {walletId: wallet.id, key},
+                      },
+                    },
+                  ],
+                }),
+              );
+            });
+          } else {
+            navigation.dispatch(
+              CommonActions.reset({
```
lanchana, line 160: question: else will execute even when getStatus return err. is it intensional?

src/navigation/wallet/WalletStack.tsx
```diff
@@ -342,6 +347,13 @@ const WalletStack = () => {
           name={WalletScreens.GLOBAL_SELECT}
           component={GlobalSelect}
         />
+        <Wallet.Screen
```
lanchana, line 350: question: how to test this?

src/store/scan/scan.effects.ts
```diff
@@ -701,3 +709,43 @@ const handlePlainAddress =
...
+    },
+  });
+};
+
+const goToJoinWallet =
+  (data: string): Effect<void> =>
+  (dispatch, getState) => {
+    console.log('Incoming-data (redirect): Code to join to a multisig wallet');
+    const keys = Object.values(getState().WALLET.keys);
+    if (keys.length === 0) {
```
lanchana, line 727: nit: can be simplified to `if (!keys.length) {`

src/navigation/wallet/WalletStack.tsx
gabrielbazan7: you need to have more than 2 backup keys and use the general scanner for the invitation code

src/navigation/wallet/screens/JoinMultisig.tsx
gabrielbazan7: Let's ignore the error in this case and just navigate to the key view

lanchana (commented):
ACK!

JohnathanWhite (approved):

PR #16, opened 2022-05-25
add copy to clipboard sending to address

lanchana:


gabrielbazan7:
Can you add the same logic to the transaction details address ? 
Cursor_and_iPhone_12

lanchana:
added copy address to tx details and also fixed txp customData to display wallet name.

gabrielbazan7 (approved):

PR #20, opened 2022-05-26
[FEAT] app migration

JohnathanWhite:
- [x] Base migration setup
- [x] UI Flow changes -> intro is only shown if migration else onboarding start
- [x] Keys (encrypted, non encrypted)
- [x] App identity
- [x] Address book
- [x] BitPay id
- [x] Migration failure modal
- [ ] Gift cards
- [x] Coinbase
- [x] Settings

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -481,6 +481,8 @@ PODS:
     - React-Core
     - SDWebImage (~> 5.11.1)
     - SDWebImageWebPCoder (~> 0.8.4)
+  - RNFS (2.20.0):
+    - React-Core
   - RNGestureHandler (2.4.1):
     - React-Core
   - RNInAppBrowser (3.6.3):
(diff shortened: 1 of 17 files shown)


JohnathanWhite:
working on the rates and balance issue

JohnathanWhite:
best way to test on IOS -> 

1. install old app and setup address book, bp id, keys, etc
2. in new app -> store/index.ts -> enable persistor.purge so that the store is cleared on every refresh
3. In xcode make sure bundle id is com.bitpay.wallet and install new app. Currently we are not clearing Cordova storage so every refresh will run migration.

JohnathanWhite:
Screen Shot 2022-05-26 at 12 36 30 PM

will have to make the above change when testing coinbase migration

gabrielbazan7 (approved):
ACK!

PR #26, opened 2022-05-27
Design updates

lanchana:


JohnathanWhite:
we need to also update the "Do more" section on home root

JohnathanWhite (approved):

PR #29, opened 2022-05-28
[FIX] decrypt before backup

JohnathanWhite:
Migrated encrypted keys are causing crashes when trying to grab mnemonic for backup. This PR adds the decrypt modal prior to navigating.


gabrielbazan7 (approved):
ACK

PR #30, opened 2022-05-28
[FIX] altCurrency settings keyboard flash

JohnathanWhite:
When selecting an altCurrency and exiting the keyboard reopens and flashes - this PR dismisses it before navigating. I also refactored and removed the on going process.


gabrielbazan7 (approved):
ACK

PR #49, opened 2022-05-31
Misc shop tab fixes and improvements

msalcala11:
1. make redeem instructions and category filter sheets scrollable
2. use FlatList for archived gift cards screen to boost performance
3. fix react-native style warning on gift card details page
4. ensure promos on shop online merchants don’t overlap with long descriptions 
5. ensure shop online merchants can handle all discount types
6. ensure swiping back from gift card details in archived view goes back to the previous screen rather than removing the whole stack
7. ensure headerTitle styling is consistent across all shop tab screens

### DIFF
--- a/src/components/modal/bottom-notification/BottomNotification.tsx
+++ b/src/components/modal/bottom-notification/BottomNotification.tsx
@@ -15,7 +15,7 @@ import {
   White,
 } from '../../../styles/colors';
 import haptic from '../../haptic-feedback/haptic';
-import {Platform} from 'react-native';
+import {Platform, ScrollView} from 'react-native';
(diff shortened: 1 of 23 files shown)


JohnathanWhite:
I've been trying to optimize the initial tab render (really slow on some android devices). Looks like moving the catalog fetch frees up the thread. Thoughts on moving the startFetchCatalog and retryGIftCardRedemptions into the app effects init? Could wrap it into a new startup effect -> startInitShop. It might add to the start up time just a bit but i think its worth it potentially as the tab load lag is bad and interactions are delayed.

JohnathanWhite (approved):

PR #62, opened 2022-06-03
Fix/login error handling

cnguyen812:
Replaced AlertBox with bottom notification modal for:

- basic login
- login 2fa authenticator/email
- forgot password
- create account

On iOS there was an issue where trying to hide the OngoingModal then showing the BottomModal was causing the BottomModal to show then hide. Seems like when the BottomModal is shown before the OngoingModal finishes the hide animation, the hide animation ends up hiding both. I used the willShow/hide Modal hooks to track the active modal in the store so we only show 1 SheetModal/OngoingProcessModal at a time.


lanchana (approved):
Thanks for the modal fix! nice work!

src/navigation/auth/screens/ForgotPassword.tsx
```diff
@@ -51,19 +49,46 @@ const ForgotPassword = () => {
...
-
-    return () => clearTimeout(timer);
-  }, [forgotPasswordEmailStatus]);
+    const {status, message} = forgotPasswordEmailStatus;
+
+    if (status === 'success') {
+      dispatch(
+        AppActions.showBottomNotificationModal({
+          type: 'success',
+          title: 'Email Sent',
```
lanchana, line 58: nit: `Email Sent` -> `Email sent`

PR #71, opened 2022-06-06
[FEAT] hide key balance

gabrielbazan7:


src/navigation/wallet/screens/KeyOverview.tsx
```diff
@@ -230,6 +239,7 @@ const KeyOverview: React.FC<KeyOverviewScreenProps> = ({navigation, route}) => {
   const defaultAltCurrency = useAppSelector(({APP}) => APP.defaultAltCurrency);
   const [showKeyDropdown, setShowKeyDropdown] = useState(false);
   const key = keys[id];
+  const {hideKeyBalance} = key;
```
lanchana, line 242: This will throw an err while deleting the key. Can be fixed in couple of ways
1. `const {hideKeyBalance} = key || {};`
2. on line no: 422 `!key?.hideKeyBalance` instead `!hideKeyBalance`

lanchana (commented):
hide it from here too? Screen Shot 2022-06-07 at 2 07 08 PM

gabrielbazan7:
Done. Thanks!

lanchana (approved):
LGTM!

PR #74, opened 2022-06-06
Send Fixes

msalcala11:
1. Check address network of scanned address before showing wallet list in global send flow
2. Ensure user-defined wallet names are shown in the wallet selector

Also fixes a typescript error related to `KeyWalletsRowProps` missing a type argument


lanchana (approved):
LGTM!

PR #83, opened 2022-06-08
Multisig Improvements

msalcala11:


src/navigation/wallet/screens/CreateMultisig.tsx
```diff
@@ -176,10 +176,13 @@ const CreateMultisig = () => {
   const dispatch = useDispatch();
   const logger = useLogger();
   const navigation = useNavigation();
+  const route = useRoute<RouteProp<WalletStackParamList, 'CreateMultisig'>>();
+  const {currency, key} = route.params || {};
+  const segwitSupported = currency !== 'BCH';
```
gabrielbazan7, line 181: Can you modified that to something like this ?
`const segwitSupported = ['btc', 'ltc'].includes(currency.toLowercase());`

msalcala11: done

JohnathanWhite (approved):

PR #91, opened 2022-06-09
Check status of multisig wallet in Key Overview before navigating

msalcala11:
Currently, creating a multisig wallet, then accepting the invite with a different device, then going back to Key Overview, and then tapping on the newly created multisig wallet takes you back to the Copayers screen even though the invite has already been accepted. This can be confusing to users, especially if they are not aware that they need to pull to refresh on the Copayers screen to complete multisig wallet creation. This PR checks the status of the newly created multisig wallet on the Key Overview screen, and, if it is complete, navigates to the Wallet Details screen rather than the Copayers screen.

### DIFF
--- a/src/navigation/wallet/screens/KeyOverview.tsx
+++ b/src/navigation/wallet/screens/KeyOverview.tsx
@@ -391,6 +391,15 @@ const KeyOverview: React.FC<KeyOverviewScreenProps> = ({navigation, route}) => {
                     // TODO
                     console.log(err);
                   }
+                  if (status.wallet.status === 'complete') {
+                    fullWalletObj.openWallet({}, () => {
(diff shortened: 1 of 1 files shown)


JohnathanWhite (approved):

src/navigation/wallet/screens/KeyOverview.tsx
```diff
@@ -391,6 +391,15 @@ const KeyOverview: React.FC<KeyOverviewScreenProps> = ({navigation, route}) => {
                     // TODO
                     console.log(err);
                   }
+                  if (status.wallet.status === 'complete') {
```
JohnathanWhite, line 394: just out of caution can we add safety operators here?

PR #113, opened 2022-06-16
Fix: removed ScrollView to avoid nested VirtualizedList

cmgustavo:
Screen Shot 2022-06-16 at 10 45 09


gabrielbazan7 (approved):
LGTM

JohnathanWhite:
was this tested with a scaled up Android phone? i think the scrollview was added to handle overflow due to sizing issues

cmgustavo:
> was this tested with a scaled up Android phone? i think the scrollview was added to handle overflow due to sizing issues

The problem is not the <scrollview> per se. The problem is that this component is in general called from inside a scrollview. The error is that there're inherit scrollviews.
Maybe we need to remove or be careful in add scrollview and call this component.

PR #137, opened 2022-06-22
Fix: getCountry for not paired users

Gamboster:
It also fixes the payment methods available in case of not getting country


src/navigation/services/buy-crypto/utils/buy-crypto-utils.ts
```diff
@@ -9,16 +9,19 @@ import {
...
   currency?: string,
   coin?: string,
 ): PaymentMethods => {
-  if (!currency || !coin || !country) {
+  if (!currency || !coin) {
     return {};
   }
-  PaymentMethodsAvailable.sepaBankTransfer.enabled = !!country.isEuCountry;
+  PaymentMethodsAvailable.sepaBankTransfer.enabled = countryData
+    ? !!countryData.isEuCountry
```
JohnathanWhite, line 23: instead of ternary could be -> !!countryData?.isEuCountry

JohnathanWhite (approved):

PR #143, opened 2022-06-24
Handle deep links on app launch

msalcala11:
Currently, deep links are only handled if the app is already open before the user taps the deep link. This means that the "Pay in Wallet" button on the BitPay invoice does not currently work if the BitPay app is not already open prior to the purchase (the most likely scenario). This PR ensures that deep links are handled even when the app is not open prior to the user tapping the deep link.

The PR also ensures line items always sum properly on the PayPro confirm screen.


cnguyen812 (approved):
LGTM

JohnathanWhite (approved):

PR #148, opened 2022-06-24
Universal link

lanchana:


src/store/scan/scan.effects.ts
```diff
@@ -196,6 +208,138 @@ const goToPayPro =
...
+        return;
+      }
+    } catch {}
+    switch (result) {
+      case 'pairingRequired':
+        navigationRef.navigate('Auth', {
+          screen: 'Login',
+          params: {
+            onLoginSuccess: () => {
+              const parentNav = navigationRef.getParent();
```
cnguyen812, line 239: This is always undefined isn't it? Since navigationRef is the root navigator?

src/store/scan/scan.effects.ts
```diff
@@ -196,6 +208,138 @@ const goToPayPro =
...
+      // needs verification - send to bitpay id verify
+      case 'userShopperNotFound':
+      case 'tierNotMet':
+        dispatch(
+          showBottomNotificationModal({
+            type: 'warning',
+            title: t('Connect Your BitPay ID'),
+            enableBackdropDismiss: false,
+            message: t(
+              'To complete this payment, please login with your BitPay ID.',
```
cnguyen812, line 262: Did these translations make it into the translation request? If not we need to queue them up for the next translations batch

src/store/scan/scan.effects.ts
```diff
@@ -196,6 +208,138 @@ const goToPayPro =
...
+    }
+  };
+
+const unlockInvoice =
+  (invoiceId: string, network: Network): Effect<Promise<string>> =>
+  async (dispatch, getState) => {
+    const {BITPAY_ID} = getState();
+    const token = BITPAY_ID.apiToken[network];
+
+    const isPaired = !!token && token !== '';
```
cnguyen812, line 293: `token !== ''` is redundant

src/store/scan/scan.effects.ts
```diff
@@ -196,6 +208,138 @@ const goToPayPro =
...
+        dispatch(showBottomNotificationModal(GeneralError()));
+        break;
+    }
+  };
+
+const unlockInvoice =
+  (invoiceId: string, network: Network): Effect<Promise<string>> =>
+  async (dispatch, getState) => {
+    const {BITPAY_ID} = getState();
+    const token = BITPAY_ID.apiToken[network];
```
cnguyen812, line 291: This may not exist, eg. app is set to mainnet but you scan a testnet invoice or vice versa

How do we want to handle this? Maybe check network !== APP.network and display an error message?

src/store/scan/scan.effects.ts
```diff
@@ -196,6 +208,138 @@ const goToPayPro =
...
+
+    try {
+      const tokens = (await BitPayIdApi.getInstance()
+        .request('getProductTokens', token)
+        .then(res => {
+          if (res.data.error) {
+            throw new Error(res.data.error);
+          }
+          return res.data;
+        })) as any;
```
cnguyen812, line 306: Can we replace `any`

src/store/scan/scan.effects.ts
cnguyen812: Following up on this, it returns `pairingRequired` but since the invoice is unclaimed it sends me to the confirm screen anyway, which may be ok? But still seems like a hole that needs to be plugged

src/store/scan/scan.effects.ts
```diff
@@ -196,6 +208,138 @@ const goToPayPro =
     }
   };
 
+const handleUnlock =
+  (data: string): Effect =>
+  async dispatch => {
+    const invoiceId = data.split('i/')[1].split('?')[0];
+    const network = data.includes('test') ? Network.testnet : Network.mainnet;
```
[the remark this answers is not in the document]
lanchana: updated to `test.bitpay.com`

src/store/scan/scan.effects.ts
lanchana: no not to the current one. yes, there is one more PR which add onGoingProcess translations, this will get into that boat.

src/store/scan/scan.effects.ts
lanchana: When it returns `pairingRequired` we make another call to check if we can fetch invoiceData 
- on success redirects to confirm screen
- on failure goes to the login/verification flow (401 if amount > 3000).

and yes we should add that check with error msg, will address that

JohnathanWhite (approved):

PR #181, opened 2022-07-01
Optimize memo component

msalcala11:
1. Fix a border styling issue when the input is both empty and in edit mode
2. Ensure taps on the save and clear buttons are handled (rather than ignored) while the keyboard is open
3. On focus, scroll sufficiently to make sure entire memo input is above the keyboard (iOS only)


cmgustavo (approved):
ACK!

PR #187, opened 2022-07-05
Feat: redirect to debug view if app init failed

cmgustavo:


package.json
```diff
@@ -99,6 +99,7 @@
     "react-native-level-fs": "3.0.1",
     "react-native-linear-gradient": "2.5.6",
     "react-native-localize": "2.1.7",
+    "react-native-mail": "6.1.1",
```
cnguyen812, line 102: Don't forget the yarn lock and pod lock files

src/Root.tsx
```diff
@@ -215,17 +215,24 @@ export default () => {
...
 
   // MAIN APP INIT
   useEffect(() => {
-    dispatch(AppEffects.startAppInit());
-  }, [dispatch]);
+    if (!failedAppInit) {
+      dispatch(AppEffects.startAppInit());
+    } else {
+      setInitFailed(true);
+      navigationRef.navigate(RootStacks.DEBUG, {name: 'Failed app init'});
```
cnguyen812, line 233: Depending where the error occurs, navRef may not be ready in the Root component yet, try navigating from the catch block in the effect

src/navigation/Debug.tsx
```diff
@@ -1,31 +1,100 @@
...
+    .map(log => {
+      const formattedLevel = LogLevel[log.level].toLowerCase();
+
+      const output = `[${log.timestamp}] [${formattedLevel}] ${log.message}\n`;
+      logStr += output;
+      return output;
+    });
+
+    const handleEmail = (data: string) => {
+      Mailer.mail({
```
cnguyen812, line 70: Where is this being mailed to? 

Should we add a disclaimer about sensitive information being included in the logs?

src/navigation/Debug.tsx
```diff
@@ -1,31 +1,100 @@
...
+  font-size: 14px;
+  color: ${Caution};
+`;
+
+const LogError = styled(BaseText)`
+  font-size: 14px;
+  color: ${({theme: {dark}}) => (dark ? '#E1E4E7' : SlateDark)};
+`;
+
 const DebugScreen: React.FC<StackScreenProps<RootStackParamList, 'Debug'>> = ({
```
cnguyen812, line 49: I think you'll need to use the useInsets hook to add some top margins, the top is getting cut off for me

cnguyen812 (dismissed):

src/Root.tsx
msalcala11: Another quick solution would be to move the `await sleep(2000);` to before the `dispatch(AppActions.failedAppInit());` to give `navigationRef` enough time to get ready (this fix worked for me no matter where the error was thrown)

msalcala11:
This is working well for me! Tested on both an iPhone X and a Galaxy S9

src/navigation/Debug.tsx
cmgustavo: Ok, I'll check. Thanks!

src/Root.tsx
cmgustavo: Yes, I agree. navRef may not be ready if error occurs early. But I prefer to add sleep before call failedAppInit since I don't like to add a navigation action into the effect.

Thank you for your suggestions!

gabrielbazan7 (approved):
LGTM

JohnathanWhite (approved):

JohnathanWhite (approved):

PR #198, opened 2022-07-07
[REF] avoid blocking app init if error

gabrielbazan7:


src/store/wallet/effects/init/init.ts
```diff
@@ -36,7 +36,6 @@ export const startWalletStoreInit =
         dispatch(
           LogActions.error(`failed [startWalletStoreInit]: ${errorStr}`),
         );
-        return reject(e);
```
cmgustavo, line 39: I thought that function was essentials to run the app. Maybe the function name is not correct, or it's not related with the stored data.
Anyway If we remove it here, it's not necessary `return resolve()`. Also it's not necessary return a Promise, and remove `await` where it function is called (`app.effect`).

cmgustavo: I don't think it block the app to start, since every function (like `getPriceHistory`) has it own try/catch.

gabrielbazan7:
https://github.com/bitpay/bitpay-app/pull/256

PR #203, opened 2022-07-08
[FIX] only save token rates if exists in tokens list

gabrielbazan7:
Avoid cannot read property `symbol` of undefined


msalcala11 (approved):
ACK. Code looks good, and runs well both in my simulator and on my device! Makes sense that only relevant rates are saved.

PR #206, opened 2022-07-09
Universal links part2

lanchana:


src/navigation/card/screens/CardPairingScreen.tsx
```diff
@@ -5,52 +5,89 @@ import {VirtualDesignCurrency} from '../../../store/card/card.types';
...
+    }
 
-    dispatch(
-      CardActions.virtualDesignCurrencyUpdated(
-        virtualDesignCurrency || 'bitpay-b',
-      ),
-    );
-  }, [dispatch, route.params?.vcd]);
+    if (paymentUrl) {
+      navigation.dispatch(StackActions.replace('Tabs', {screen: 'Home'}));
```
cnguyen812, line 44: Since Home is always open, I feel like performance would be better by just navigating instead of replacing. Otherwise it seems like it's unmounting everything then remounting Tabs/Home again. Maybe by doing a check to see if Tabs/Home exists in the stack somewhere first, else do a replace?

JohnathanWhite (approved):

cnguyen812 (approved):

PR #210, opened 2022-07-11
Ref: remove unused functions

cmgustavo:
* Also adds required params to Coinbase Balance Card


msalcala11 (approved):
ACK. Great to remove unused imports, and add all useMemo/useEffect deps

PR #215, opened 2022-07-11
Fix botched OfferCard merge

msalcala11:
Currently causes duplicate navigation when tapping on an OfferCard


Gamboster (approved):
ACK

JohnathanWhite (approved):

PR #216, opened 2022-07-11
Fix (Coinbase): check if token is invalid and suggest re-connect

cmgustavo:


msalcala11 (approved):
ACK! I was prompted to reconnect when my token was invalid.

PR #223, opened 2022-07-12
Fix: avoid select signed txp. If multipleSignAvailable is active...

cmgustavo:
* If multipleSignAvailable is active, do not open details view, just select clicking on the row. 
* Adds SafeAreaView

Simulator Screen Shot - iPhone 12 - 2022-07-12 at 16 31 11


msalcala11 (approved):
ACK

PR #227, opened 2022-07-12
Fix: avoid keyboard open above the recovery phrase input

cmgustavo:
Related to: https://github.com/bitpay/bitpay-app/pull/181

### DIFF
--- a/src/navigation/wallet/components/RecoveryPhrase.tsx
+++ b/src/navigation/wallet/components/RecoveryPhrase.tsx
@@ -12,6 +12,7 @@ import {
   ActiveOpacity,
   CtaContainer as _CtaContainer,
   HeaderContainer,
+  ScreenGutter,
 } from '../../../components/styled/Containers';
 import Button from '../../../components/button/Button';
 import {useDispatch, useSelector} from 'react-redux';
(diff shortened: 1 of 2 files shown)


src/navigation/wallet/components/RecoveryPhrase.tsx
```diff
@@ -557,8 +561,10 @@ const RecoveryPhrase = () => {
   }, []);
 
   return (
-    <ScrollViewContainer>
-      <ImportContainer>
+    <ScrollViewContainer
+      extraScrollHeight={50}
```
msalcala11, line 565: Since the recovery phrase input is larger than the memo input, more `extraScrollHeight` is needed to ensure the input is fully above the keyboard. Seems like `90` is an optimal value (works well on an iPhone SE)

msalcala11 (changes requested):

src/navigation/wallet/components/RecoveryPhrase.tsx
cmgustavo: Great! It looks better. 

Thanks!

msalcala11 (approved):

PR #230, opened 2022-07-13
Add Testnet wallet buy - faucet links

lanchana:
- Testnet Wallet buy: faucet links
- ~Import wallet: increase input text font size, reduce spacing~
- ~Wallet connect: Display import wallet notification when no compatible wallets available~


src/navigation/wallet/screens/WalletDetails.tsx
```diff
@@ -947,34 +955,43 @@ const WalletDetails: React.FC<WalletDetailsScreenProps> = ({route}) => {
...
+                              },
                             },
-                            opts: {
-                              hideSendMax: true,
-                              context: 'buyCrypto',
-                            },
-                          },
-                        });
+                          });
+                        } else {
```
Gamboster, line 992: I would put a warning before leaving the app, saying something like that because it is a testnet wallet you cannot make a purchase, however you will be taken to an external faucet where you can claim test coins.

lanchana: good idea! will add that

JohnathanWhite:
merge conflicts

src/navigation/wallet/screens/WalletDetails.tsx
Gamboster: I was rather suggesting something like this:
iPhone_12
What do you think?

In addition, let's make sure the faucets are trusted. I would not put redirects to websites that could be a source of scams against our users.

gabrielbazan7:
I'm not really into this idea. What about just showing a suggestion to the user ? something like: You choose a testnet wallet. You can get free coins in searching a btc testnet faucet.

src/navigation/wallet/screens/WalletDetails.tsx
lanchana: added the notification

lanchana: Checking with @JohnathanWhite about faucet links

PR #232, opened 2022-07-14
Fix home memo cards language update

lanchana:


src/navigation/card/screens/settings/Referral.tsx
```diff
@@ -184,7 +184,7 @@ const Referral = ({}) => {
 
   const getStatus = (status: string, expiration: number) => {
     return status === 'pending' && currentDate >= expiration
-      ? 'Expired'
+      ? t('Expired')
       : status.charAt(0).toUpperCase() + status.slice(1);
```
cnguyen812, line 188: Is status enumerated somewhere? If so we should create a map and map each known status to a translated value

src/constants/LanguageSelectionList.tsx
```diff
@@ -5,38 +5,47 @@ interface ItemProps {
 
 export const LanguageList: Array<ItemProps> = [
   {
+    // t('Dutch')
```
cnguyen812, line 8: Why no translate?

PR #233, opened 2022-07-14
Fix: enable renew subscription for silent push notifications

cmgustavo:
* Renew subscription for receive silent push notification and keep app active when it's open. 
* Debounce functions to avoid multiple calls
* Update list of transactions proposals when receive an update
* Ignore warning: Settings a time for long period of time (https://github.com/ably/ably-js/issues/749#issuecomment-455680414)


gabrielbazan7 (approved):
ACK!!

PR #234, opened 2022-07-14
fix - swap init

JohnathanWhite:


msalcala11 (approved):
ACK. Code looks good and works well. I'm able to select a swap destination wallet.

PR #240, opened 2022-07-14
Feat/braze event listener

cnguyen812:
Add event listener for Braze content card updates. Instead of just fetching content cards directly, we request a refresh and on refresh event, then we fetch content cards. 

- on init, subscribe a listener and persist the subscription to listen for updates, and fetch cards on update event
- on init, since it takes a bit of time for content cards to update, we poll a few times at 5 second intervals for initial content cards
- request refresh on HomeRoot focus, throttled to every 10 seconds
- update Braze language on language change


cmgustavo (approved):
LGTM

JohnathanWhite (approved):

PR #244, opened 2022-07-15
Gift card improvements

msalcala11:
1. Ensures that only a single tap is required to navigate to a gift card search result when the keyboard is open (currently two taps are required, with the first tap closing the keyboard)
2. Fixes a wonky stack reset that happens when purchasing a gift card with Coinbase by ensuring that the Coinbase 2fa screen is part of the GiftCardStack so that all screens in the checkout flow are all part of the same stack and therefore dismissed simultaneously


lanchana (approved):
LGTM!

PR #247, opened 2022-07-18
Scroll to top when tapping on the active tab.

lanchana:


cmgustavo:
IMO I prefer to keep the position where I left the "cursor". It helps for example when you are checking balance before buy a gift card from last merchant in the list

gabrielbazan7:
I agree with Gustavo. The way to do this is double tap on the tab icon to scroll to top.

PR #251, opened 2022-07-19
Use memo component in Transaction Details

msalcala11:
Also, this PR allows users to clear the memo entirely after it has been set (currently not supported) and ensures that the updated memo is instantly reflected in the parent wallet details view after being changed in the child transaction details view (without needing to manually pull to refresh to see the change)


gabrielbazan7 (approved):
ACK

cmgustavo (approved):
LGTM

PR #260, opened 2022-07-20
Feat/validation translations

cnguyen812:
Added support for translated default yup validation messages:

- From now on, import `yup` from the lib folder instead of from node_modules
- Set the default validation message for `required`, `email`, `min`, `max` to a translation function. Add more as needed
- Re-exported yup since setLocale must be run before creating any schemas.
  - Schemas defined outside of components are run in unknown order, so just did it this way instead of moving all schemas into components
  - If the schema overrides the default message, moved the schema into the component so it language will change properly
- added JSX interpolated translation to CreateAccount TOU label

For interpolated translations, the value will be `lorem ipsum <0>{name}</0> and <1>{id}</1> lorem ipsum` where `name` and `id` will be values and `<0/>` and `<1 />` are interpolated components. Use the `<Trans />` tag with the values and components props to pass in the data. Useful for injecting links and such into translations (see CreateAccount for real example).

Usage:
```
translation.json
{
  "sampleKey": "lorem ipsum <0>{name}</0> and <1>{id}</1> lorem ipsum"
}

Example.tsx
<Trans key="sampleKey" values={{ name: 'satoshi', id: 1 }} components={[<Text />, <Anchor />]} />

Output:
lorem ipsum <Text>satoshi</Text> and <Anchor>5</Anchor> lorem ipsum
```


locales/en/translation.json
```diff
@@ -1131,5 +1128,17 @@
...
+  "Required": "Required",
+  "CannotUseSameCodeTwice": "Cannot use the same code twice.",
+  "EnterArgArgArgToActivateYourCard": "Enter {{0}}, {{1}}, and {{2}} to activate your card.",
+  "EnterArgArgToActivateYourCard": "Enter {{0}} and {{1}} to activate your card.",
+  "EnterArgToActivateYourCard": "Enter {{0}} to activate your card.",
+  "EnterYourInformationToActivateYourCard": "Enter your information to activate your card.",
+  "Segwit": "Segwit",
+  "Testnet": "Testnet",
+  "InvalidInvitationCode": "Invalid invitation code",
+  "MustBeLessThanArgCharacters": "Must be less than {{0}} characters",
```
cnguyen812, line 1141: Since this isn't a list, it doesn't have to be `0` I guess, it could be `max` or something, might update this in a bit

src/lib/yup/index.ts
```diff
@@ -0,0 +1,22 @@
...
+    email: () => t('Please enter a valid email address.'),
+    min: params => t('MustBeGreaterThanArgCharacters', {0: params.min}),
+    max: params => t('MustBeLessThanArgCharacters', {0: params.max}),
+  },
+});
+
+/**
+ * Re-export yup from here so that setLocale is run before creating any schemas.
+ */
+export default yup;
```
cnguyen812, line 22: Definitely open to suggestions for better approaches

src/navigation/wallet/screens/CreateMultisig.tsx
```diff
@@ -546,7 +541,7 @@ const CreateMultisig = () => {
                     });
                   }}>
                   <Column>
-                    <OptionTitle>Segwit</OptionTitle>
+                    <OptionTitle>{t('Segwit')}</OptionTitle>
```
gabrielbazan7, line 544: Are we sure we want to translate this ?

src/navigation/wallet/screens/CreateMultisig.tsx
```diff
@@ -566,7 +561,7 @@ const CreateMultisig = () => {
               <AdvancedOptions>
                 <RowContainer onPress={toggleTestnet}>
                   <Column>
-                    <OptionTitle>Testnet</OptionTitle>
+                    <OptionTitle>{t('Testnet')}</OptionTitle>
```
gabrielbazan7, line 564: Same here

src/navigation/wallet/screens/CreateMultisig.tsx
cnguyen812: Not sure, seemed like a missed strings. I can remove

gabrielbazan7: Yes, thanks !

gabrielbazan7 (approved):
nice! ACK

PR #269, opened 2022-07-25
contact row updates

rastajpa:


rastajpa:
https://github.com/bitpay/bitpay-app/pull/268

PR #274, opened 2022-07-26
Fix phrase overflow

lanchana:


cmgustavo (approved):
LGTM! Tested on XS (iOS) and S9 (Android)

PR #275, opened 2022-07-26
update contacts row and cleanup

lanchana:


src/navigation/wallet/components/SendToContact.tsx
```diff
@@ -138,7 +140,7 @@ const SendToContact = () => {
             {contacts.map((item, index) => {
               return (
                 <SendContactRow key={index}>
```
rastajpa, line 142: Could you remove this container (SendContactRow), even delete it because it is no longer used in the whole app and move the `key={index}` to ContactRow. This is because unnecessary top and bottom padding appear in the view.

rastajpa (approved):
LGTM!

PR #283, opened 2022-07-27
option to switch between networks

lanchana:


src/navigation/tabs/settings/general/screens/Theme.tsx
```diff
@@ -14,11 +19,76 @@ import {AppActions} from '../../../../../store/app';
...
+    const _clickCount = clickCount + 1;
+    setClickCount(_clickCount);
+    if (_clickCount >= 10) {
+      const changeNetwork =
+        network === Network.mainnet ? Network.testnet : Network.mainnet;
+
+      dispatch(
+        showBottomNotificationModal({
+          type: 'info',
+          title: `${t('Test Mode ')}${
```
cnguyen812, line 51: Don't concatenate partial sentence translations, use interpolation instead

tsx
`t(Test Mode )t(Enabled)` => `t((TestModeArg, {status: t(Enabled)})`

translation.json
`"Test Mode ": "Test Mode "` =>  `"TestModeArg": "Test Mode {{status}}"`

This way in case the order is different in another language (eg. JP, AR, etc) it translates better

src/navigation/tabs/settings/general/screens/Theme.tsx
```diff
@@ -14,11 +19,76 @@ import {AppActions} from '../../../../../store/app';
...
+          ),
+          enableBackdropDismiss: false,
+          actions: [
+            {
+              text: t('CONTINUE'),
+              primary: true,
+              action: async () => {
+                dispatch(networkChanged(changeNetwork));
+                await sleep(200);
+                RNRestart.Restart();
```
cnguyen812, line 65: Is a restart necessary? Why not just update the store? App should be using the store network as SSOT

src/navigation/tabs/settings/general/screens/Theme.tsx
```diff
@@ -14,11 +19,76 @@ import {AppActions} from '../../../../../store/app';
...
+    navigation.setOptions({
+      headerTitle: () => {
+        return (
+          <Pressable onPress={onPressTitle}>
+            <HeaderTitle>{t('Theme')}</HeaderTitle>
+          </Pressable>
+        );
+      },
+    });
+  });
```
cnguyen812, line 90: Updating navigation options is expensive, we should add dependencies to useLayoutEffect

src/navigation/tabs/settings/general/screens/Theme.tsx
cnguyen812, line 51: Do we even need translations for this? Isn't this more for internal/debug use?

JohnathanWhite: yeah dont need translations

src/navigation/tabs/settings/general/screens/Theme.tsx
```diff
@@ -14,11 +19,75 @@ import {AppActions} from '../../../../../store/app';
...
+    navigation.setOptions({
+      headerTitle: () => {
+        return (
+          <Pressable onPress={onPressTitle}>
+            <HeaderTitle>{t('Theme')}</HeaderTitle>
+          </Pressable>
+        );
+      },
+    });
+  }, [navigation, t, onPressTitle]);
```
cnguyen812, line 89: onPressTitle is going to make this run every render as well, for situations like this I usually throw the event handler into a ref, it's kinda clunky but performant
```
const onPressTitle = () => {
  // logic
};
const onPressTitleRef = useRef(onPressTitle);
onPressTitleRef.current = onPressTitle;

useLayoutEffect(() => {
  // call onPressTitleRef.current() here
}, []);
```

src/navigation/tabs/settings/general/screens/Theme.tsx
```diff
@@ -14,11 +19,75 @@ import {AppActions} from '../../../../../store/app';
...
+    const _clickCount = clickCount + 1;
+    setClickCount(_clickCount);
+    if (_clickCount >= 10) {
+      const changeNetwork =
+        network === Network.mainnet ? Network.testnet : Network.mainnet;
+
+      dispatch(
+        showBottomNotificationModal({
+          type: 'info',
+          title: `Test Mode ${
```
cnguyen812, line 51: Maybe change the word order to like `Enable Test Mode` and `Disable Test Mode` instead of `Test Mode Enabled` (past tense)

src/navigation/tabs/settings/general/screens/Theme.tsx
lanchana: oh nice, ty! updated it.

cnguyen812 (approved):

PR #325, opened 2022-08-04
Fix: swap checkout publish and sign

Gamboster:
This portion of code was removed here by mistake apparently. So the swaps are not currently working: https://github.com/bitpay/bitpay-app/pull/303/files#diff-75567b38dc413b88bf0bfcd2f225e5bb12ffc0c2bc349405197b4b6cc2fe1f74L502


rastajpa (approved):
ACK

PR #331, opened 2022-08-06
Fix: email notifications

cmgustavo:
* Fix: Unsubscribe email notifications from Settings
* Subscribe new created/imported wallets to email notifications if enabled
* Unsubscribe push/email notification when key is deleted


gabrielbazan7 (approved):
LGTM!

src/store/wallet/effects/create-multisig/create-multisig.ts
```diff
@@ -29,6 +37,19 @@ export const startCreateKeyMultisig =
...
+        // subscribe new wallet to email notifications
+        if (
+          emailNotifications &&
+          emailNotifications.accepted &&
+          emailNotifications.email
+        ) {
+          const prefs = {
+            email: emailNotifications.email,
+            language: defaultLanguage,
+            unit: 'btc', // deprecated
```
gabrielbazan7, line 49: Since unit is deprecated I think we can remove it

PR #333, opened 2022-08-08
Feat/polygon UI

cnguyen812:
Refactor currency selection to support new UI. Tokens for a specific chain currency are now selected on a separate screen, with the exception of popular tokens. 

To test:
Create new basic wallet (multi select)
Create new multisig wallet (single select)
Add new basic wallet to existing key (single select)
Add new multisig wallet to existing key (single select)

Screen Shot 2022-08-08 at 1 08 10 PM


gabrielbazan7 (approved):
ACK!

PR #338, opened 2022-08-09
Clear wallet transaction history before saving keys to device storage

msalcala11:
Currently, wallet transactions are persisted to storage despite being wiped again after refetching keys in the outbound redux-persist transform (which bloats storage unnecessarily). This PR ensures transactions are never persisted to device storage.


cmgustavo (approved):
LGTM

PR #339, opened 2022-08-09
Refactor rate storage

msalcala11:
Moves rates out of the `WALLET` storage key to a new dedicated `RATE` key to ensure that only the minimum amount of data is stored in the `WALLET` key to avoid running into OS-specific per-key storage limits.


cmgustavo:
This looks great! Nice refactor.

JohnathanWhite (approved):

PR #351, opened 2022-08-11
Fix: prevent swipe back action returns to Home tab

cmgustavo:
* Go to Create -> New Key -> Swipe back (it goes to Home instead `Select an option` screen)

After fix:

* Swipe back is disabled on the `Select Currencies` screen.


gabrielbazan7 (approved):
ACK

JohnathanWhite (approved):

PR #381, opened 2022-09-07
Fix/add physical card

cnguyen812:
This PR adds the `Order Physical Card` link in card settings.

Testing:
Should ONLY be visible if no valid physical card exists. 
Should open the IAB (NOT the WebView, so should see Chrome/Safari custom tabs within the app) and request authentication
After authenticating, should be redirected to the order physical flow as usual


JohnathanWhite (approved):

JohnathanWhite:
Screen Shot 2022-09-07 at 3 47 21 PM


https://bitpay.zendesk.com/agent/tickets/434584
https://bitpay.zendesk.com/agent/tickets/433952

PR #392, opened 2022-09-09
[REF] dont remove txp if err - hide checkbox from already signed txp

gabrielbazan7:
Do not show sign checkbox from already accepted txps


cmgustavo (approved):
ACK! 👍🏼

PR #411, opened 2022-09-15
[RN-661]: Adding crypto address with scanning closes the contacts screen

rastajpa:
on android devices 500 ms debounce time was not enough to avoid wrong behaviors


gabrielbazan7 (approved):
ACK! testnet on android

PR #419, opened 2022-09-20
Fix: added required "android:exported" for new targetSdkVersion (31)

cmgustavo:
Getting error:

```
Manifest merger failed : Apps targeting Android 12 and higher are 
required to specify an explicit value for `android:exported` when 
the corresponding component has an intent filter defined. 
See https://developer.android.com/guide/topics/manifest/activity-element#exported for details.
```


gabrielbazan7 (approved):
ACK!

Gamboster (approved):

PR #421, opened 2022-09-21
Ref: adds ActivityIndicator to avoid delay when setting language

cmgustavo:
https://user-images.githubusercontent.com/237435/191552134-302226c9-a881-4ae0-bdef-3c4f209991ff.mp4


gabrielbazan7 (approved):
Nice

PR #430, opened 2022-10-03
[REF] simplify adding new chains - 01

gabrielbazan7:
Fixes global selector
Order wallets by priority
Improve variable names


Gamboster (approved):
ACK

PR #435, opened 2022-10-06
Fix: importing a first wallet on new installation asks to accept TermsofUsers twice

cmgustavo:
After accept terms when importing a wallet and back to Home view. Waiting for the import finish... Click on "ok", instead to open the new imported key, it asks to accept terms again.

ERROR:

https://user-images.githubusercontent.com/237435/194414872-178e633d-057f-4a57-b8b6-aee70ab66b3c.mov

FIXED:

https://user-images.githubusercontent.com/237435/194414117-7ce4bd19-c0ed-4bf6-9cc6-555182f0ae6a.mov


rastajpa (approved):
LGTM

PR #437, opened 2022-10-07
Feat/braze custom attr referral

cnguyen812:
This MR adds a custom Braze plugin to the Segment SDK to support some cloud-mode features, particularly for anonymous users and to limit our identify calls.

Additionally it sends the user's referral code as the `Cardholder Unique Referral Code` attribute to be stored in Braze.

Refactored a bit of the way we identify users due to how we are using EIDs for anonymous users and also to not track the user's account if they have not authorized tracking.

New ENV variables required:
`SEGMENT_API_KEY_ANDROID`
`SEGMENT_API_KEY_IOS`
`BRAZE_REST_API_ENDPOINT`: Braze REST API endpoint so we can make queries
`BRAZE_EXPORT_API_KEY`: this API key is for fetching the `brazeId`, in Braze this key should  be configured for `users.export.ids`

Deprecated ENV variables:
`SEGMENT_API_KEY`: split into Android/iOS

Message me and I can show you where to get the keys.


cmgustavo (approved):
LGTM! Nice work! 🚀

PR #442, opened 2022-10-11
[FIX] getReceivingAddressChain function

gabrielbazan7:


src/navigation/bitpay-id/utils/bitpay-id-utils.ts
```diff
@@ -1,9 +1,9 @@
+import {SUPPORTED_COINS} from '../../../constants/currencies';
 import {ReceivingAddress} from '../../../store/bitpay-id/bitpay-id.models';
-import {IsERCToken} from '../../../store/wallet/utils/currency';
 
 export function getReceivingAddressChain({
   chain,
   currency,
 }: ReceivingAddress): string {
-  return chain || (IsERCToken(currency) ? 'eth' : currency);
+  return chain || (!SUPPORTED_COINS.includes(currency) ? 'eth' : currency);
```
msalcala11, line 8: just realized we should do `currency.toLowerCase()` here and we should make 'eth' capitalized to match the convention of how receiving addresses are stored in the database

msalcala11: Rolled this change into this PR: https://github.com/bitpay/bitpay-app/pull/444

PR #460, opened 2022-10-18
Fix Contacts crash

msalcala11:
Old contacts saved in storage currently don't have a `chain` field (this field was added as part of the matic refactor). This PR ensures backwards compatibility with the old contacts schema (fixing a crash on app startup for users who have contacts without a `chain` field set)

### DIFF
--- a/src/utils/helper-methods.ts
+++ b/src/utils/helper-methods.ts
@@ -320,9 +320,9 @@ export const addTokenChainSuffix = (name: string, chain: string) => {
   return `${name.toLowerCase()}_${chain.charAt(0)}`;
 };
 
-export const getCurrencyAbbreviation = (name: string, chain: string) => {
-  return IsERCToken(name.toLowerCase(), chain.toLowerCase())
-    ? addTokenChainSuffix(name, chain)
+export const getCurrencyAbbreviation = (name: string, chain?: string) => {
+  return IsERCToken(name, chain)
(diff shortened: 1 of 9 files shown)


cmgustavo (dismissed):
LGTM!

src/utils/helper-methods.ts
```diff
@@ -320,9 +320,9 @@ export const addTokenChainSuffix = (name: string, chain: string) => {
   return `${name.toLowerCase()}_${chain.charAt(0)}`;
 };
 
-export const getCurrencyAbbreviation = (name: string, chain: string) => {
-  return IsERCToken(name.toLowerCase(), chain.toLowerCase())
-    ? addTokenChainSuffix(name, chain)
+export const getCurrencyAbbreviation = (name: string, chain?: string) => {
+  return IsERCToken(name, chain)
+    ? addTokenChainSuffix(name, chain!)
```
rastajpa, line 325: if chain is optional this condition could break some case

rastajpa (changes requested):

src/utils/helper-methods.ts
msalcala11: Which case would it break? In `IsERCToken` we check to make sure chain is defined before returning `true`, so that addTokenChainSuffix always has a defined chain passed in to it.

PR #468, opened 2022-10-20
Fix value copied to clipboard when tapping qr code for enabling 2fa

msalcala11:
Also makes email search only possible for logged in users for now.


cmgustavo (approved):
LGTM

PR #470, opened 2022-10-21
[FIX] add tokens to preferences

gabrielbazan7:
Tokens added from the `add wallet` option were not added to the wallet preferences, so when importing those tokens they were missing


rastajpa (approved):
nice!

PR #474, opened 2022-10-24
Ref: swap crypto update wallet status

Gamboster:


rastajpa (approved):
I like that balance info modal

PR #502, opened 2022-10-26
Fix: remove unnecessary code

Gamboster:
- WalletSelectorModal.tsx is no longer necessary as it was replaced.
- Removed imports and unused variables


rastajpa (approved):
LGTM

PR #509, opened 2022-10-27
[FEAT]: WalletConnect wallet_switchEthereumChain

rastajpa:


src/navigation/wallet-connect/components/Connections.tsx
```diff
@@ -47,7 +47,8 @@ export default ({
   wallet: Wallet;
   connectors: IWCConnector[];
 }) => {
-  const {walletName, network, chain} = wallet.credentials;
+  const {walletName, credentials} = wallet;
```
gabrielbazan7, line 50: it should not be necessary to extract `credentials` with the recent changes in the wallet transform function.

src/navigation/wallet-connect/screens/WalletConnectHome.tsx
```diff
@@ -103,21 +105,49 @@ const WalletConnectHome = () => {
...
   });
 
+  const [requests, setRequests] = useState<IWCRequest[]>();
+
+  const allKeys = useAppSelector(({WALLET}: RootState) => WALLET.keys);
+
+  const decodeData = async () => {
+    const promises = [];
+    for await (const r of _requests) {
+      const data = r.payload.params[0].data;
```
escottalexander, line 119: @rastajpa Are you sure the method you are passing here is one that is unique to the MRC20 ABI? If not, there is a possibility that it will return as ERC20 as they share 7 methods in common.

rastajpa: good catch! you're right, I've been testing and it's not possible, so this workaround will fail in some cases. I'm working on a new solution, thank you

rastajpa: thanks @escottalexander, now I sent some changes with a better way to get chain. Changes in bitcore will not be necessary for the moment.

rastajpa:
https://github.com/bitpay/bitpay-app/pull/608

PR #515, opened 2022-10-28
Feat: handle buy limits in Amount view

Gamboster:
In this PR:
- Handle buy limits in Amount view
- Fix: reset real amount value when switch between currencies (Without this fix, you can enter a number, switch between currencies in the Amount view and the view will show 0, however the "Continue" button will still be enabled, since the actual value remains as it was)
- Fix: Added a warning message when trying to enter an amount bigger than max limit
- Fix: added translations to warning messages
- Fix: context: 'buyCrypto' from walletDetails and CardDashboard


gabrielbazan7 (dismissed):
LGTM

gabrielbazan7 (approved):

PR #516, opened 2022-10-28
Revert "[FIX] get contact name function"

JohnathanWhite:
Reverts bitpay/bitpay-app#513


msalcala11 (commented):
ACK. I can confirm reverting this change fixes the "Could not update transaction history" error for me

msalcala11 (approved):

PR #518, opened 2022-10-28
Fix: check if custom token and show historic rates

cmgustavo:
### Currently issue

For example if we pass btc to the function `IsCustomERCToken`. 
The `currency` variable returns `btc_b`.

The next function BitpaySupportedCurrencies[`btc_b`] returns `false` ... and finally it's `!`, so it returns `true`. 

Finally the function `IsCustomERCToken` will return `true` for `BTC`

Screen Shot 2022-10-28 at 09 51 22

FIXED

Screen Shot 2022-10-28 at 09 54 48

### DIFF
--- a/src/store/wallet/utils/currency.ts
+++ b/src/store/wallet/utils/currency.ts
@@ -67,7 +67,7 @@ export const IsCustomERCToken = (
   currencyAbbreviation: string,
   chain: string,
 ) => {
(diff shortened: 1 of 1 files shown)


gabrielbazan7 (approved):
good catch. ACK

Gamboster (approved):
ACK

PR #525, opened 2022-10-31
Fix Coinbase: BCH deposit not working due to new format BCH address

cmgustavo:
Now Coinbase generates new BCH format address. 
So, it's not necessary to convert it from old format to new one, because bitcore returns error trying to do this function with a BCH address: 

```
Bitcore.Address(addressToTranslate)
```


gabrielbazan7 (approved):
ACK

PR #528, opened 2022-11-01
[REF] better handle of unsent transactions

gabrielbazan7:
Include again all types of transaction proposals in badge count

new `Unsent Transactions` header that includes read only wallets txs (no sign available for this) and 1-1 transactions that need a sign ( created by a readonly wallet in another device)

iPhone_13

Failed paypro transactions are listed in `Rejected` ( no sign available for this )

iPhone_13

iPhone_13


Gamboster (approved):
ACK

PR #549, opened 2022-11-11
[REF] token selection

gabrielbazan7:
- show number of assets supported
- show number of wallets to be created
- keep blue button in tokens view
- Separate core BitPay tokens from others
- `Select Ethereum Tokens` instead of `Select Ethereum Currencies`

Old:
Cursor

New:

Cursor_and_Slack


rastajpa (dismissed):
LGTM

JohnathanWhite (approved):

PR #550, opened 2022-11-14
Minor fixes for running in iOS

v1nsai:
* Fixes a typedef redefinition error related to `RCT-Folly`.  [Details in this thread.](https://github.com/facebook/flipper/issues/834#issuecomment-905774601)
* Adds the `ios-deploy` library for deploying to device, as suggested by `yarn ios:device` when I tried without it.


package.json
```diff
@@ -70,6 +70,7 @@
     "https-browserify": "1.0.0",
     "i18next": "21.6.4",
     "intl": "1.2.5",
+    "ios-deploy": "^1.11.4",
```
cmgustavo, line 73: This package should be on the `devDependencies` instead.

PR #573, opened 2022-11-30
[RN-770]: fix fiat rate str in txp view 

rastajpa:
IMG_A48EA0E61F2A-1_jpeg

and missing hr in settings...
iPhone_13


cmgustavo (approved):
ACK! Good catch.

PR #588, opened 2022-12-13
Backwards compatible deeplinks

cnguyen812:
With the AppsFlyer SDK in place, we can now handle OneLink links, however any embedded deeplink inside a OneLink does not get handled by the Navigation Container and the linking configuration we have configured. Added logic so that if our custom handling doesn't handle the passed in URL, then default to the Navigation Container deeplinking.


cmgustavo (approved):
ACK! 👍🏼

JohnathanWhite (approved):

PR #589, opened 2022-12-14
[RN-781]: ios/android quick links

rastajpa:


src/constants/shortcuts.ts
```diff
@@ -0,0 +1,36 @@
+import {Platform} from 'react-native';
+import {ShortcutItem} from 'react-native-quick-actions';
+
+export const ShortcutList: Array<ShortcutItem> = [
+  {
+    type: 'buy',
+    title: 'Buy Cryptos',
```
cmgustavo, line 7: Just a typo: "Buy Crypto"
Screenshot 2022-12-16 at 13 27 26

cmgustavo (approved):
Nice work! Tested on iOS and Android
LGTM 👍🏼

cmgustavo (changes requested):
I found an issue specifically on Android. The event `quickActionShortcut` is permanently called every time you enter (or back) to Home. So, if you click on a shortcut function (receive, send, buy, or swap), then you go back to Home, it will return to the view "Select a currency". Also, if you go back, and change the tab, it's fine. But if you return to Home, it will redirect again to "Select a currency".

src/constants/shortcuts.ts
rastajpa: ouch! ty!

rastajpa:
> I found an issue specifically on Android. The event `quickActionShortcut` is permanently called every time you enter (or back) to Home. So, if you click on a shortcut function (receive, send, buy, or swap), then you go back to Home, it will return to the view "Select a currency". Also, if you go back, and change the tab, it's fine. But if you return to Home, it will redirect again to "Select a currency".

thank you, I'm working on a fix

cmgustavo (dismissed):
Worked on Android! 👍🏼

cmgustavo (approved):
ACK! 🚀

Gamboster (approved):
LGTM
Nice work!!

PR #593, opened 2022-12-15
FIX [RN-786]: translations and interpolation issues

cmgustavo:
* Fix "OnGoingProcessMessages". i18next can't translate `enum` values. Best solution was to move it as `type` and `object` with translate function.
* Use string param for ´startOnGoingProcessMessage` instead `enum`.
* Stop using OnGoingProcessMessage directly. Moved to ´startOnGoingProcessMessage`.
* Updated some strings for es.json file.
* Fixed all warnings `Key is not a string literal` when using translation:extract


rastajpa:
LGTM!

rastajpa (approved):
LGTM!

PR #599, opened 2022-12-20
Fix[RN-785]: autoselect first testnet wallet if env is DEV. Show testnet badge

cmgustavo:
Screenshot 2022-12-20 at 16 32 34

### DIFF
--- a/src/navigation/services/buy-crypto/screens/BuyCryptoRoot.tsx
+++ b/src/navigation/services/buy-crypto/screens/BuyCryptoRoot.tsx
@@ -27,7 +27,6 @@ import {CurrencyImage} from '../../../../components/currency-image/CurrencyImage
 import {RootState} from '../../../../store';
 import {
   showBottomNotificationModal,
-  dismissBottomNotificationModal,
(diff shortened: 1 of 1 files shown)


Gamboster:
I like the idea of adding the Testnet badge 👍🏼 

Regarding the selection of the wallet, JFI I did not implement it this way because some exchanges request livenet addresses despite being in a sandbox environment (for example, Simplex) and others (such as Wyre) require a testnet address to test in the sandbox environment.

Whether or not we make the change, the dev will often have to hand-pick the wallet anyway to ensure they can properly test a particular exchange.

cmgustavo:
> I like the idea of adding the Testnet badge 👍🏼
> 
> Regarding the selection of the wallet, JFI I did not implement it this way because some exchanges request livenet addresses despite being in a sandbox environment (for example, Simplex) and others (such as Wyre) require a testnet address to test in the sandbox environment.
> 
> Whether or not we make the change, the dev will often have to hand-pick the wallet anyway to ensure they can properly test a particular exchange.

I agree. Closed for now.

PR #606, opened 2022-12-27
[FEAT] react-native-screens freeze

JohnathanWhite:
This PR enables freeze mode for react-native-screens which freezes the subtree for screens that are not visible, improving performance. 

https://github.com/software-mansion/react-freeze#quick-start-with-react-navigation-react-native-

One way to test the performance gain is to try switching theme with the enableFreeze flag true and false.


Gamboster (approved):
ACK! 👏🏼

PR #607, opened 2022-12-27
[REF] add imported wallet to created key if already exist

gabrielbazan7:


msalcala11 (commented):
ACK! Works great for manually importing a duplicated BCH wallet via a specified derivation path and attaching it to the existing key

msalcala11 (approved):

Gamboster (approved):
ACK!

PR #622, opened 2023-01-06
Removes Wyre integration from supported buy crypto exchanges

Gamboster:
I didn’t remove everything related to Wyre. Only the things that I considered necessary. I left the code related to old Wyre transactions, because I think that could be useful for some users. The list of transactions only will be visible for users who completed any purchase in the past with Wyre.  I also added a warning message for them saying that this exchange is out of service now:

```
          title: t('Warning'),
          message: t('Please note that Wyre is out of service.'),
          type: 'warning',
```


cmgustavo (approved):
LGTM!

PR #636, opened 2023-01-16
Feat/granular tracking updates

cnguyen812:
This PR refactors how we handle the tracking permission request. Functionality is mostly the same and if the user requests no tracking, we do not enable the Segment SDK or any of its plugins, and also ignore any calls to identify/track/screen. However we do call `changeUser` with Braze with **only** the EID and no personal information like email, name, etc. Associating the EID with the device (and device push token) enables Braze/BWS/BitPay platform to send push notifications to the correct device.


cmgustavo (approved):
ACK! Nice work 👍🏼

PR #644, opened 2023-01-20
[REF] handle wallet_does_not_exist err correctly

gabrielbazan7:
To trigger this error enter the 12 words phrase and set an incorrect passphrase in the advanced option


Gamboster (approved):
ACK

PR #654, opened 2023-01-31
[FIX] Signup params

JohnathanWhite:


JohnathanWhite:
testing

should be able to create an account:
- with marketing opt in enabled
- disabled

cmgustavo (approved):
ACK. Worked to me 👍🏼

PR #660, opened 2023-02-08
Fix japanese translation

middlingphys:


cmgustavo:
Translations should be performed in Crowdin: https://crowdin.com/project/copay/ja

PR #681, opened 2023-02-16
Feat: Ramp Integration

Gamboster:
This PR needs: https://github.com/bitpay/bitcore/pull/3542

The Ethereum network does not work in the Sandbox environment (neither ETH nor its Tokens), it will give you an error if you want to buy ETH. According to the list of "supported currencies", it is supported (https://docs.ramp.network/assets). We will have to test if it works correctly with the production credentials.

Currently it allows you to edit the destination address in their checkout page. But according to this answer in their FAQ (https://ramp.notion.site/ramp/Integration-FAQ-3208ceccaa5648ac9b9dc9cbf8d61289#a586006682ee4a818266920b491a23c4) we must ask Ramp team to block the possibility of editing this field for greater control and to prevent users from making mistakes:

> ```The API key allows you to customize and track your integration and access our newest features. [Fill out this form](https://share.hsforms.com/1TKWT1ksNQZ6AURRV0ssvqg5awa0) and get full access to Ramp's SDK. In case you need more support, contact us at [partner@ramp.network](mailto:partner@ramp.network):> ```
> ```lock crypto address - if you'd like to prevent User changing the crypto address, we can lock it for you> ```

Some particularities of the Ramp integration to take into account:

GetQuote from Ramp does not allow consultation with country code, so if the user arrives at the checkout page, they may find limits different from those shown in the app (usually lower) that are more linked to the selected payment method

Once you've made a purchase with Ramp, it saves your previous selections, and the user's previous selections take precedence over choices they make later in the app. This can be noticed with the payment method and fiat currency selected.

Tests in sandbox environment:
If you want to advance until completing the payment in Sandbox you will have to use test addresses. So you can test it with a BTC testnet wallet.

The payment method to test in this environment is "Manual Bank Transfer" :

```
You can make a test purchase without having to connect your own payment method (staging environments only):
1. Add a new payment method, select "Manual Bank Transfer" (switch country to any European country if your country doesn't have that option).
2. You should be back to the confirmation screen with "Any {currency} account" manual bank transfer payment method selected.
3. Continue, the purchase will be created.
4. Tick the box that says that you've transferred the funds and continue.
5. Open the transaction summary link.
6. Click the manual test release button at the bottom of the summary and wait a few moments until the confirmation is processed.
```


rastajpa (approved):
LGTM. Tested with local BWS/BWC and it works good. Nice work

JohnathanWhite (approved):

PR #685, opened 2023-02-21
Improve invoice web url scan handling

msalcala11:
Currently invoice web url scan handling only works if `id` is the first query param in the url. This PR ensures scan handling works regardless of the order the query params in the web url.


gabrielbazan7 (approved):
ACK

PR #687, opened 2023-02-21
Fix: show fiat value for livenet tx on Confirm screen

cmgustavo:
Fixes https://github.com/bitpay/bitpay-app/pull/605


gabrielbazan7 (approved):
ACK

PR #690, opened 2023-02-23
[RN-857]: export transaction history

rastajpa:


src/navigation/wallet/screens/wallet-settings/ExportTransactionHistory.tsx
```diff
@@ -0,0 +1,291 @@
...
+        title: csvFilename,
+        url: `file://${filePath}`,
+        subject: `${walletName} Transaction History`,
+      };
+
+      await RNFS.writeFile(filePath, csv, 'utf8');
+
+      if (option === 'download') {
+        await Share.open(opts);
+      } else {
```
cmgustavo, line 225: I'd change this part with a native mail plugin `react-native-mail` in order to check if the user has defined default email client. Also we could hide if it's not available.

cmgustavo (approved):
LGTM!

src/navigation/wallet/screens/wallet-settings/ExportTransactionHistory.tsx
rastajpa: nice! I'll make the changes, ty!

PR #692, opened 2023-02-24
Feat: global error handler to capture uncaught exceptions. Adds new event

cmgustavo:
https://github.com/a7ul/react-native-exception-handler 

Persist log for exceptions after the app crash.

Screenshot 2023-03-14 at 15 28 39

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,10 +2,14 @@ import 'react-native-get-random-values'; // must import before @ethersproject/sh
 import '@ethersproject/shims';
 import './shim';
 import '@walletconnect/react-native-compat';
-import {AppRegistry} from 'react-native';
+import {AppRegistry, Alert} from 'react-native';
 import Root from './src/Root';
 import React from 'react';
 import './i18n';
+import {
+  setJSExceptionHandler,
+  setNativeExceptionHandler,
+} from 'react-native-exception-handler';
 import {name as appName} from './app.json';
 import getStore from './src/store';
 import {Provider} from 'react-redux';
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,16 +1,42 @@
...
 import 'react-native-url-polyfill/auto'; // https://github.com/facebook/react-native/issues/23922#issuecomment-648096619
 import {enableFreeze} from 'react-native-screens';
+import { Analytics } from './src/store/analytics/analytics.effects';
+import { APP_VERSION } from './src/constants/config';
+import { GIT_COMMIT_HASH } from '@env';
 enableFreeze(true);
 
+const errorHandler = (e, isFatal) => {
+  if (isFatal) {
+    Analytics.track('BitPay App - Crashed App', { build: GIT_COMMIT_HASH, version: APP_VERSION});
```
cnguyen812, line 20: Should be: 
```
store.dispatch(Analytics.track(...));
```

and moved after the `store` declaration.

index.js
```diff
@@ -1,16 +1,42 @@
...
+      [{
+        text: 'Close'
+      }]
+    );
+  }
+};
+setJSExceptionHandler(errorHandler, true);
+
+const exceptionHandler = () => {
+    Analytics.track('BitPay App - Crashed App', { build: GIT_COMMIT_HASH, version: APP_VERSION });
```
cnguyen812, line 36: Should be: 
```
store.dispatch(Analytics.track(...));
```

and moved after the `store` declaration.

package.json
```diff
@@ -95,6 +95,7 @@
     "react-native-device-info": "8.4.7",
     "react-native-draggable-flatlist": "3.0.7",
     "react-native-error-boundary": "1.1.12",
+    "react-native-exception-handler": "2.10.10",
```
cnguyen812, line 98: Don't forget the Podfile.lock

index.js
cnguyen812: Do we want to track the actual error as well? Or add to persistent logs?

cnguyen812 (changes requested):

index.js
cmgustavo: Yeah, good idea. I'll take a look. Thanks!

index.js
cmgustavo: 👍🏼

index.js
```diff
@@ -1,21 +1,66 @@
...
+const errorHandler = (e, isFatal) => {
+  if (isFatal) {
+    store.dispatch(
+      Analytics.track('BitPay App - Crashed App', {
+        build: GIT_COMMIT_HASH,
+        version: APP_VERSION,
+      }),
+    );
+    const errStr = e instanceof Error ? e.message : JSON.stringify(e);
+    LogActions.persistLog(LogActions.error(errStr));
```
cnguyen812, line 36: This needs to be dispatched as an store action

```
store.dispatch(LogActions...)
```

cnguyen812 (dismissed):

index.js
cmgustavo: Yes, fixed. Thanks!

index.js
```diff
@@ -2,22 +2,67 @@ import 'react-native-get-random-values'; // must import before @ethersproject/sh
 import '@ethersproject/shims';
 import './shim';
 import '@walletconnect/react-native-compat';
-import {AppRegistry} from 'react-native';
+import {AppRegistry, Alert} from 'react-native';
+import 'react-native-get-random-values';
```
msalcala11, line 6: Why is this import needed?

cmgustavo: Oh... it's duplicated (from line 1). Good catch!

msalcala11 (approved):
Works great!

PR #711, opened 2023-03-10
Fix/deeplinks and in app messaging

cnguyen812:
This PR always initializes the Segment SDK so that AppsFlyer deeplinks are always handled. The app tracking request for the user is handled by the Segment IDFA plugin, so the react-native-tracking-transparency dependency has been removed. ID for advertising continues to be not collected if user declines. 

Should resolve issues on some iOS devices where OneLink links not redirecting properly.


cmgustavo (dismissed):
LGTM 👍🏼

JohnathanWhite (approved):

PR #719, opened 2023-03-21
[FEAT] pre-commit sensitive information warning file

gabrielbazan7:
By default, Git looks for hooks in the `.git/hooks` directory, which is not version-controlled. To tell Git to use the custom `.githooks` directory instead, run the following command in the terminal: 

```
git config core.hooksPath .githooks
```

example: 

```
> git commit -m "..."            
WARNING: You are attempting to commit changes which include a 'no-commit'. Make sure you are not including sensitive information.
Please check the following files:
   - android/app/src/main/AndroidManifest.xml
   - android/app/src/main/res/values/braze.xml
   - ios/BitPayApp/AppDelegate.m

You can ignore this warning by running the commit command with '--no-verify'
```

Also I added the `ALLOWED-URL-PREFIXES-REPLACE-ME` values since there is no sensitive information there, it's always the same and it's always needed for running the app.
Same for `<base-config cleartextTrafficPermitted="true" />` always needed for running the app in dev mode

### DIFF
--- a/android/app/src/main/res/xml/network_security_config.xml
+++ b/android/app/src/main/res/xml/network_security_config.xml
@@ -1,7 +1,7 @@
 <network-security-config>
   <!-- Set cleartextTrafficPermitted = true for development -->
-  <base-config cleartextTrafficPermitted="false" />
-  <domain-config cleartextTrafficPermitted="false">
+  <base-config cleartextTrafficPermitted="true" />
+  <domain-config cleartextTrafficPermitted="true">
     <domain includeSubdomains="true">localhost</domain>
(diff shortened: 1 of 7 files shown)


cmgustavo (dismissed):
Nice!

android/app/src/main/res/xml/network_security_config.xml
```diff
@@ -1,7 +1,7 @@
 <network-security-config>
   <!-- Set cleartextTrafficPermitted = true for development -->
-  <base-config cleartextTrafficPermitted="false" />
```
JohnathanWhite, line 3: can we revert this to false

PR #727, opened 2023-03-23
Ref: Simplex logo and icon

Gamboster:
Simplex has updated their logos and icons, as seen in this link:
https://www.simplex.com/simplex-brand-assets


cmgustavo (approved):
ACK

PR #735, opened 2023-03-29
[FIX] check for stored multisig proposals

gabrielbazan7:
iPhone_14_Pro

To reproduce: 

1. Create a transaction proposal
2. Close the app and remove the transaction with the other copayer
3. Open the app and go to wallet details with the first copayer


cmgustavo (approved):
LGTM!

PR #737, opened 2023-03-30
[REF] store index instead of React Element for uiIcons

gabrielbazan7:
Need help for testing amazon giftcard


cmgustavo (dismissed):
ACK! 👍🏼

msalcala11 (approved):
ACK. Confirmed that gift card icons are showing up properly in transaction history

PR #741, opened 2023-03-30
[FIX] erc20 pending proposals view

gabrielbazan7:
For showing the matic_eth wallet in the header we need bws changes. But, I think this way actually makes sense and lgtm ( showing the associated eth wallet )

iPhone_14_Pro


rastajpa (approved):
LGTM

PR #742, opened 2023-03-31
Fix: Moonpay SEPA bank transfer getQuote

Gamboster:
Moonpay only accepts EUR as a base currency for SEPA payments.
This PR forces to use EUR if the selected fiat currency of the app is other than EUR.


rastajpa (approved):
ACK

PR #762, opened 2023-04-13
Fix: prevent multiple redir after link Coinbase account (only Android)

cmgustavo:
It also happens with any deeplinks (BitPay Card link, etc)

https://stackoverflow.com/questions/45676629/how-can-we-remove-deep-link-data-from-intent


gabrielbazan7 (approved):
ACK

PR #765, opened 2023-04-18
Ref: TXP Details

Gamboster:
txs renamed to txp
re-broadcast TXP feature added
Check PayPro details added
Missing status messages added
Show "Spendable balance modal" when click on "Total Locked Balance" row
Fixed "danger" links buttons to show the color correctly


cmgustavo (approved):
LGTM!

PR #768, opened 2023-04-20
[REF] change default for RBF flag

gabrielbazan7:
Since we are creating the transaction proposal before moving to the confirm view ( the only view that had the RBF option ) now checking for the `enableReplaceByFee` flag is needed in `createProposalAndBuildTxDetails`


src/navigation/wallet/screens/send/confirm/Confirm.tsx
```diff
@@ -142,7 +142,7 @@ const Confirm = () => {
   const [showPaymentSentModal, setShowPaymentSentModal] = useState(false);
   const [resetSwipeButton, setResetSwipeButton] = useState(false);
   const [showTransactionLevel, setShowTransactionLevel] = useState(false);
-  const [enableRBF, setEnableRBF] = useState(false);
```
JohnathanWhite, line 145: this should default to the app settings level flag

JohnathanWhite (approved):

PR #764, opened 2023-04-14
Ref: check threshold for Coinbase payments. Show message if no wallet available. Single request for Invoice data

cmgustavo:
* Do not show Coinbase wallets if it's not enabled for pay BitPay Invoice
* Removes duplicated http request for BitPay Invoice
* Show generic message if there're no wallets for selected BitPay Invoice

PR #787, opened 2023-05-02
Fix: do not show incorrect message if there are wallets

cmgustavo:
Fix an introduced error here: https://github.com/bitpay/bitpay-app/pull/764/files#diff-22411585442c1a32be5b8f0aecb10594a5917bd6dfe076c44173cd71264e45d9R570


gabrielbazan7 (approved):
ACK

PR #805, opened 2023-05-16
Fix: disallowing screen capture when app in background

cmgustavo:


cmgustavo:
No necessary since it's defined when window is created: https://github.com/bitpay/bitpay-app/pull/805/files#diff-0bf9de0e1eebe5a0e48017e40e38f4e188872383c727787e999900a6d82629dfR50

PR #846, opened 2023-07-07
Feat: check if email verified on Profile screen

cmgustavo:
Show "Receive Email address" and "Two-Factor Authentication" options only if email has been verified.


gabrielbazan7 (approved):
LGTM

PR #848, opened 2023-07-10
[RN-977]: In-app notifications

rastajpa:
(WalletConnect implementation only)

iPhone_14_Pro_and_ar_-_BitPay_-_Slack


gabrielbazan7 (approved):
ACK

PR #853, opened 2023-07-12
Walletconnect v2 notification adjustments

JohnathanWhite:
- adjusted timing
- adjusted backdrop
- removed reset
- added context to auto open confirm if from notification
- reversed requests list


cmgustavo (approved):
ACK!

PR #868, opened 2023-07-20
Fix/warnings

cnguyen812:
This PR fixes the following warnings:

```
WARN  'new NativeEventEmitter()' was called with a non-null argument without the required 'addListener' method.
WARN  'new NativeEventEmitter()' was called with a non-null argument without the required 'removeListeners' method.
```

- upgraded `react-native-appboy-sdk` from `1.38.1` => `1.40.0` and changed `kotlin_version` to `kotlinVersion` per their release notes
- added stub methods to the Android GooglePushProvisioning java module
- pass `null` into event emitter if module not defined for that OS

```
WARNING: This environment is missing a secure random source; generated private keys may be at risk, think VERY carefully about not adding a better secure source.
```

- rearranged imports so that `react-native-get-random-values` is imported before `@ethersproject/shims`

### DIFF
--- a/android/app/src/main/java/com/bitpay/wallet/GooglePushProvisioningModule.java
+++ b/android/app/src/main/java/com/bitpay/wallet/GooglePushProvisioningModule.java
@@ -45,6 +45,16 @@ public class GooglePushProvisioningModule extends ReactContextBaseJavaModule {
   }
 
   @ReactMethod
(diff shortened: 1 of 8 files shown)


JohnathanWhite:
@cnguyen812 can you regenerate the pod lock?

JohnathanWhite (approved):

PR #869, opened 2023-07-21
Fix: add removing bill label

msalcala11:


msalcala11:
included here: https://github.com/bitpay/bitpay-app/pull/872

PR #875, opened 2023-07-26
Ensure non-null value passed to NativeEventEmitter

cnguyen812:
This PR fixes the error where null is passed to the NativeEventEmitter ctor.


cmgustavo (approved):
LGTM!

PR #880, opened 2023-08-01
Feat: show fiat amounts in select inputs view

Gamboster:
- Show fiat amounts in select inputs view: I added the fiat values of each input and the total, in the same selection view. Currently this is not displayed and for certain use cases it is cumbersome, since the only time you can see the fiat equivalents is in the final confirm view
- Some minor UI fixes

Notification_Center


gabrielbazan7 (approved):
nice. LGTM !!

PR #889, opened 2023-08-11
[FIX] oneinch api

gabrielbazan7:


src/store/wallet/effects/currencies/currencies.ts
```diff
@@ -26,11 +27,17 @@ export const startGetTokenOptions =
       let tokenOptionsByAddress: {[key in string]: Token} = {};
       let tokenData: {[key in string]: CurrencyOpts} = {};
       for await (const chain of SUPPORTED_EVM_COINS) {
+        const url = `https://api.1inch.dev/swap/v5.2/${EVM_BLOCKCHAIN_ID[chain]}/tokens`;
```
cmgustavo, line 30: Should we move it to `constant`? Easier to change in a future.

JohnathanWhite (approved):

PR #899, opened 2023-08-16
Fix/log dispatch

cnguyen812:
This PR fixes a few persistent logs that fail to dispatch, and fixes the home screen wallet carousel height.


gabrielbazan7 (approved):
Good catch. ACK

PR #906, opened 2023-08-24
[FIX] confirm PIN section

gabrielbazan7:
The "confirm pin" when setting a pin view was skipped. This PR will fix that and also I refactor some part of the code to avoid unnecessary re renders.


Gamboster (approved):
LGTM

PR #907, opened 2023-08-24
Fix/recovery phrase carousel

cnguyen812:
This PR fixes the recovery/verify phrase screens where performance was choppy and the words were not responding on the verify screen. Refactored to prevent `useLayoutEffect` from running every render, as well as not creating new `renderItem` functions every render.


gabrielbazan7:
I'm running into an issue: `TypeError: Object is not a function` ( RecoveryPhrase.tsx )

gabrielbazan7 (approved):
ACK!

PR #909, opened 2023-08-25
[REF] Store persisted logs in a different entry in storage

gabrielbazan7:


msalcala11 (approved):
ACK. Confirmed persisted logs work properly with this new storage location.

PR #917, opened 2023-08-31
[FIX] send

gabrielbazan7:


msalcala11 (approved):
ACK. Confirmed that this fixes an error during the token send flow.

cmgustavo (approved):
ACK

PR #932, opened 2023-09-21
Fix: removed momentum scroll that doesn't work on DesktopApp

cmgustavo:
Known issue: https://github.com/facebook/react-native/issues/38090


src/navigation/wallet/screens/WalletDetails.tsx
```diff
@@ -1195,15 +1193,9 @@ const WalletDetails: React.FC<WalletDetailsScreenProps> = ({route}) => {
         ItemSeparatorComponent={() => <BorderBottom />}
         ListFooterComponent={listFooterComponent}
         onEndReached={() => {
-          if (!onEndReachedCalledDuringLoading) {
-            loadHistory();
-            setOnEndReachedCalledDuringLoading(true);
-          }
-        }}
-        onEndReachedThreshold={0.5}
```
JohnathanWhite, line 1203: was it intentional to change threshold from 0.5 to .01?

cmgustavo: Thanks for review! It wasn't intentional. I was testing different values and forgot to keep in 0.5 in order to load at the half of elements.

JohnathanWhite (approved):

PR #939, opened 2023-09-26
Ref: remove clipboard functions from SendTo screen

cmgustavo:


gabrielbazan7:
Nice. Totally agree. Being asked for permission to read the clipboard every time is quite annoying. We can still paste manually.

PR #941, opened 2023-09-27
[FIX] rate calculation when creating using crypto amount instead of usd

gabrielbazan7:
Choosing any other currency than usd here https://bitpay.com/demos/ results in incorrect rate calculations

### DIFF
--- a/src/store/wallet/effects/send/send.ts
+++ b/src/store/wallet/effects/send/send.ts
@@ -421,17 +421,17 @@ export const buildTxDetails =
       fee = proposal.fee || 0; // proposal fee is zero for coinbase
     }
 
-    const invoiceCurrency =
+    const selectedTransactionCurrency =
       invoice?.buyerProvidedInfo!.selectedTransactionCurrency ||
       wallet.currencyAbbreviation.toUpperCase();
 
     const isOffChain = !proposal;
-    if (invoice && invoiceCurrency) {
+    if (invoice && selectedTransactionCurrency) {
       amount = isOffChain
-        ? invoice.paymentSubtotals[invoiceCurrency]
-        : invoice.paymentTotals[invoiceCurrency];
+        ? invoice.paymentSubtotals[selectedTransactionCurrency]
+        : invoice.paymentTotals[selectedTransactionCurrency];
       const coinAndChain = getCoinAndChainFromCurrencyCode(
-        invoiceCurrency.toLowerCase(),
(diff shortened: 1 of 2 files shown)


Gamboster (dismissed):
ACK

cmgustavo (dismissed):
ACK! 👍🏼

src/store/wallet/effects/send/send.ts
```diff
@@ -360,7 +360,7 @@ export const getInvoiceEffectiveRate =
     const invoiceCurrency = getCurrencyCodeFromCoinAndChain(coin, chain);
     return (
       precision &&
-      invoice.price /
+      invoice.usdAmount /
```
msalcala11, line 363: this change will break the effective rate calculation for other fiat currencies like EUR, GBP, etc and is not needed

src/store/wallet/effects/send/send.ts
```diff
@@ -445,11 +445,12 @@ export const buildTxDetails =
     }
 
     amount = Number(amount); // Support BN (use number instead string only for view)
-    const effectiveRate =
-      (invoice &&
-        invoiceCurrency &&
-        dispatch(getInvoiceEffectiveRate(invoice, invoiceCurrency, chain))) ||
-      undefined;
+    let effectiveRate;
+    if (invoice && invoiceCurrency && defaultAltCurrencyIsoCode === 'USD') {
```
msalcala11, line 449: This change will skip the effective rate calculation for other currencies like EUR, GBP, etc. Instead we can do: `defaultAltCurrencyIsoCode === invoice.currency` and this PR will then fix the crypto pricing currency issue while keeping the effective rate properly calculating for other fiat currencies

msalcala11 (changes requested):

src/store/wallet/effects/send/send.ts
gabrielbazan7: Thanks for reviewing. Using only `defaultAltCurrencyIsoCode === invoice.currency` will skip the effective rate calculation if for example `btc` is the currency selected when creating the invoice, however we could calculate the effective rate in that case because we have `usdAmount` in the invoice properties. Tell me if the changes make sense to you.

msalcala11 (approved):

msalcala11:
ACK! Tested the latest commit on a variety of scenarios, and this PR handles all of them well.

PR #948, opened 2023-10-04
Fix: dismiss SheetModal on background mode

cmgustavo:


msalcala11 (approved):
ACK! Fixes the blurred screen issue when returning to the app after it had a modal open in the background and the PIN screen is enabled.

PR #984, opened 2023-11-20
Improve app launch speed

msalcala11:
Ensures that app launch init logic is not blocked by any `await` tasks.


gabrielbazan7 (approved):
LGTM

PR #994, opened 2023-12-04
[REF] move to native stack navigation

gabrielbazan7:


Gamboster:
I've been testing the navigation a lot and it seems to work well. The only issue I detect is that the `TransactMenu` doesn't close after select an option. The app freezes.

gabrielbazan7:
@Gamboster fixed. Thanks for reviewing

Gamboster (dismissed):
LGTM 
It seems to be working correctly, I tested everything I could in the iOS simulator. The issue I mentioned previously was fixed.

gabrielbazan7:
https://github.com/bitpay/bitpay-app/pull/1005

PR #996, opened 2023-12-07
[REF] handle contract addresses rates in batches up to 5

gabrielbazan7:
{"timestamp":"2023-12-07T12:58:33.505+00:00","error_code":10012,"status":{"error_message":"Number of contract addresses in the request exceeds the allowed limit (10 addresses). Please note that to ensure fair usage and prevent abuse, the allowed limit may be adjusted periodically. To enjoy higher allowed limit, please subscribe to Demo plan or a paid plan at https://www.coingecko.com/en/api/pricing"}}


cmgustavo (approved):
LGTM

PR #1026, opened 2024-01-12
Adjust claim logic

lanchana:


src/store/scan/scan.effects.ts
```diff
@@ -325,6 +324,12 @@ const handleUnlock =
         return;
       }
     } catch {}
+
+    if (context === 'u') {
```
JohnathanWhite, line 328: can you leave some comments on all of these context variables? "u", "c" etc

JohnathanWhite (approved):

PR #1027, opened 2024-01-13
Allow any BTC tx to be RBF'd (full RBF)

kajoseph:
I've tested this locally. Just add `mempoolfullrbf=1` to your Bitcoin Core config.

Note: This also fixes a bug where deselecting RBF won't set `txp.enableRBF` to false. Not really necessary with full rbf, but nice to have code working correctly anyway :)


gabrielbazan7 (approved):
LGTM

PR #1034, opened 2024-01-22
Fix: deeplinks handling when lock method is enabled

Gamboster:
How to reproduce the error:

iOS:
Enable the PIN/Biometric lock method:
- If the app is closed (Not in the background) and you click on a deeplink: There is a race between the lock modal and the navigation to the destination page. If the modal is showed before the navigation, then the app crashes and needs to restart.

Proposed solution: Wait for the lock modal to be successfully dismissed before performing the navigation.

- If the app is in the background there are 3 possible cases to take into account:
1) It has been minimized more than 20 seconds ago since the last unlock (`lockAuthorizedUntil`). So when you click on a deeplink, you return to the app, it shows the lock modal and a case similar to the previous one happens, in which a race occurs between the lock modal and the link navigation.

2) It has been minimized after the selected lock method has been successfully dismissed, and the deeplink has been clicked within 20 seconds (lockAuthorizedUntil), so the lock modal should not be displayed and only navigation should be performed.

3) The app has been minimized with the lock modal already open. In this case the lock modal should continue to be shown and wait for it to be correctly dismissed to continue with the link

Proposed solution: Control the `lockAuthorizedUntil` variable, to verify whether the time has been met or not. If the time is not met, the deeplink navigation is carried out normally. If the time has already passed or the modal was previously open, wait for the lock modal to be successfully dismissed before performing the navigation.

Android:
After some testing, it seems to work correctly, so no changes are necessary here.

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -16,7 +16,6 @@ import {
   Linking,
   NativeEventEmitter,
   NativeModules,
-  Platform,
   StatusBar,
 } from 'react-native';
 import 'react-native-gesture-handler';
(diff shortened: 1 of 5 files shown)


cmgustavo (approved):
ACK! 👍🏼

PR #1040, opened 2024-02-01
Ref: New Zenledger API

cmgustavo:
* Create Token if it doesn't exist and Check if it expired only when using "CreatePortfolio".
* Create Portfolio: send wallet addresses, name, etc and return aggcode and url.


Gamboster (approved):
ACK. Tested on iOS simulator

PR #1041, opened 2024-02-02
Fix: Transak "Other" payment method

Gamboster:
With this fix, the user will be able to correctly view all the payment methods enabled for the user's selections (in the Transak web checkout)


gabrielbazan7 (approved):
ACK

PR #1044, opened 2024-02-06
Fix: Renew Braze EID after logout from BitPayID

cmgustavo:
If no renew EID after logout from BitPay, I'm going to continue seeing In-App messages, Cards, etc, specifically for a BitPay User.


gabrielbazan7 (approved):
LGTM

PR #1055, opened 2024-02-14
Fix: save selected Home Layout using the app in foreign language

cmgustavo:
* Currently the Home Layout (Carousel or List view) only works in English.
* Smoother goBack function after click on Save Layout.


gabrielbazan7 (approved):
ACK

PR #1064, opened 2024-02-21
Fix: build iOS for new InAppMessageHandler

cmgustavo:
`InAppMessageHandle` file not found when building the app in production mode.

Moved both files to root `ios` folder and added to Xcode Project.

Rename `AppDelegate.mm` to `AppDelegate.m`

See: https://github.com/bitpay/bitpay-app/pull/1033


gabrielbazan7 (approved):
ACK

PR #1070, opened 2024-02-26
Ref: Banxa payment methods

Gamboster:
With these changes, the Banxa offer should display faster on the offers page.
It also includes a refactor necessary for an upcoming Banxa update regarding payment methods.


gabrielbazan7 (approved):
ACK

PR #1078, opened 2024-03-04
Fix initial tab selection in CustomizeHome.tsx

msalcala11:
The Tab Navigator can be a bit unreliable when it comes to programmatically selecting an initial nested tab that isn't the first tab. This PR includes a workaround for this issue by ensuring that the initially selected tab is always listed first in `CustomizeHome.tsx`.


cmgustavo (approved):
ACK!

PR #1100, opened 2024-03-22
Fix: prevent unlock modal from being displayed twice in a row when the app is open

Gamboster:
Way to reproduce the issue: Leave the app open for more than 20 seconds, in which case the `lockAuthorizedUntil` will have expired. When closing and reopening the app, the value of `lockAuthorizedUntil` was not removed, so after successfully passing the unlock to open the app, it interprets that `lockAuthorizedUntil` has expired and asks for it again.

The solution I propose is to add `lockAuthorizedUntil` to the `appReduxPersistBlackList`


cmgustavo (approved):
ACK!

PR #1115, opened 2024-04-02
[REF] biometric plugin

gabrielbazan7:


msalcala11 (approved):
Works great on iOS and Android!

PR #1126, opened 2024-04-12
Fix: renew Braze EID for (silent) push notifications

cmgustavo:
It helps to force subscription to (silent) push notifications after login or logout BitPay ID


gabrielbazan7 (approved):
ACK ! good catch

PR #1136, opened 2024-04-19
Fix: Sell Crypto: handle wallets without funds

Gamboster:
- Fix the app stuck error when having wallets without funds in Sell Crypto flow.
- Shows a "Buy Crypto" button that takes the user through Buy crypto flow, if they have wallets but no funds.
- Fix: Transac menu: Sell is shown disabled like the other send options when there are no funds.
- Added PaymentMethod and User id to the URL to open the Sell checkout webpage
- Fix: prevent app from crashing when quickly clicking on backdrop in wallet modals selectors


cmgustavo (approved):
ACK 👍🏼

PR #1138, opened 2024-04-22
Remove "Pay Bills" sub tab from shop tab

msalcala11:
Now that users have had enough time to get used to the location of the new top-level bill pay tab, we can remove the "Pay Bills" sub tab from the shop tab.


cmgustavo (approved):
LGTM!

PR #1142, opened 2024-04-24
[FEAT] List currencies and group them by network + Search and chain selector components

gabrielbazan7:
- Search with network Filter added to Key View - Wallet Connect Connections View - Key Settings - Global Selector
- Global select improvements: List coins - Show available network on the right side + amount of wallets available


msalcala11 (approved):
Nice work!

PR #1154, opened 2024-05-07
Prevent home screen jump on app launch

msalcala11:


gabrielbazan7 (approved):
ACK. No jumps or glitches observed on the Android version.

PR #1155, opened 2024-05-08
[FIX] Adjust the height of the chain selector modal

gabrielbazan7:


msalcala11 (approved):
Nice! Chain selector stays above the keyboard now even when the filtered list of chains is very short.

PR #1172, opened 2024-05-22
[FIX] small fixes for new chains

gabrielbazan7:


src/constants/SupportedCurrencyOptions.tsx
```diff
@@ -262,7 +262,7 @@ export const SupportedEvmCurrencyOptions: Array<SupportedCurrencyOption> = [
     id: Math.random().toString(),
     img: CurrencyListIcons.eth,
     priority: 5,
-    currencyName: 'Ethreum Arbitrum',
+    currencyName: 'Ethereum Arbitrum',
```
JohnathanWhite, line 265: why are we adding this? to identify native chain currency?

gabrielbazan7: Yes. The ETH on Arbitrum is essentially the same ETH that you use on the Ethereum mainnet, bridged to the Arbitrum network and is the coin used to pay transaction fees.

JohnathanWhite: ahh I see - studying up on it now

JohnathanWhite (approved):

PR #1191, opened 2024-06-05
Fix: get coin and chain from symbols with multiple underscores

Gamboster:
Example: usdc_c_e

Coin: usdc_c
Chain: eth


gabrielbazan7 (approved):
LGTM

PR #1195, opened 2024-06-10
Fix: zero amount in stablecoin transactions

cmgustavo:


gabrielbazan7 (approved):
ACK. Tested with usdc and eth wallet

PR #1200, opened 2024-06-14
Fix: position for search-component on KeySettings

cmgustavo:
Alignment to Design.

Screen Shot 2024-06-14 at 16 31 08


gabrielbazan7 (approved):
ACK

PR #1212, opened 2024-06-24
[FIX] avoid header flicker

gabrielbazan7:


msalcala11 (approved):
Fixes all flicker for me on iOS 👌

PR #1223, opened 2024-06-27
Fix gift card claim link sharing on android

msalcala11:
The `url` share property is only supported on iOS, which is currently making the "Share claim code" gift card feature useless on android. This PR falls back to sharing gift card claim links via the `message` property on android.


cmgustavo (approved):
LGTM

PR #1241, opened 2024-07-03
Revert: Fix wallet name when creating new key for new chains

cmgustavo:
Revert: https://github.com/bitpay/bitpay-app/pull/1224


gabrielbazan7 (approved):
ACK

PR #1260, opened 2024-07-15
Fix: show badge on Confirm View for SendTo field

cmgustavo:
Original

Simulator Screen Shot - iPhone 14 - 2024-07-15 at 16 33 00

Fixed:

Simulator Screen Shot - iPhone 14 - 2024-07-15 at 16 44 42


gabrielbazan7 (approved):
ACK good catch

PR #1264, opened 2024-07-19
Improve price chart performance

msalcala11:
1. Makes charts load faster
2. Makes switching between time ranges faster
3. Animates chart when switching between time ranges


cmgustavo (dismissed):
ACK! Nice work! 👍🏼

JohnathanWhite (approved):

PR #1274, opened 2024-08-01
Default Android navigation animation

JohnathanWhite:
Switching to default animation as it feels snappier and more performant


cmgustavo (approved):
LGTM

PR #1304, opened 2024-08-21
Fix: Coinbase and GlobalSelector

cmgustavo:
* Changed old modal for new GlobalSelector modal
* Ref: GlobalSelector to filter wallets for modals


gabrielbazan7 (approved):
ACK !

PR #1314, opened 2024-08-27
[REF] Filter options based on available chains

gabrielbazan7:
This update changes the filter to show only the options that match the items in the list. For example, if the list has BTC and LTC wallets, the filter will only show BTC and LTC as options.


cmgustavo (approved):
LGTM

PR #1321, opened 2024-08-29
Feat: auto-detect and create token wallets with funds

Gamboster:


src/store/wallet/effects/create/create.ts
```diff
@@ -728,3 +743,94 @@ export const getDecryptPassword =
...
+
+export const detectAndCreateTokensForEachEvmWallet =
+  ({key, force}: {key: Key; force?: boolean}): Effect<Promise<void>> =>
+  async dispatch => {
+    try {
+      dispatch(
+        LogActions.info('starting [detectAndCreateTokensForEachEvmWallet]'),
+      );
+
+      const evmWalletsToCheck = key.wallets.filter(
```
gabrielbazan7, line 755: you can use: `!IsERCToken(currencyAbbreviation, chain) && IsEVMChain(chain)` for checking if it is a gas wallet

`const evmWalletsToCheck = key.wallets.filter((w)=>!IsERCToken(currencyAbbreviation, chain) && IsEVMChain(chain));`

src/store/wallet/effects/create/create.ts
```diff
@@ -728,3 +743,94 @@ export const getDecryptPassword =
...
+              options: {
+                network: Network.mainnet,
+                ...(account !== undefined && {
+                  account,
+                  customAccount,
+                }),
+              },
+            };
+
+            await dispatch(
```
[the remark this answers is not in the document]
Gamboster: In this case I need the promises to be executed in sequential order and not in parallel, otherwise there is a race in the editing and saving of data within key.

src/store/wallet/effects/create/create.ts
Gamboster: done

JohnathanWhite (approved):

PR #1365, opened 2024-09-20
Fix: key updates when there are changes in settings

Gamboster:
Fixes: 
- The placeholder is not auto-filling in the Account name change form.
- When a new Key name is saved, it is not updated where names are displayed in settings, for example in the "Export wallet" option


gabrielbazan7 (approved):
LGTM

PR #1483, opened 2024-11-28
Feat: PayPal and Venmo added as Payment method opts

Gamboster:
In this PR:
- PayPal and Venmo added as supported paymentMethods/withdrawalMethods for Moonpay (Buy/Sell)
- New order for paymentMethods/withdrawalMethods in the selector
- Ref: PaymentMethod models renamed to WithdrawalMethod in sell crypto
- Fix: Fixed an issue where Sell crypto for EU countries only allows EUR as fiatCurrency
- Fix: Small UI fix for Keys selector when list is large (Bottom key not fully displayed)
- Fix: Now paymentMethod is added to moonpay-url so that it is preselected when reaching their web checkout.


cmgustavo (approved):
ACK!

PR #1499, opened 2024-12-10
Fix: TestID using old bottom sheet

cmgustavo:
* Also add TestID for provider name (Crypto Offers View)

Key-options

Offers


JohnathanWhite (approved):

Classicrichblog (commented):
Approve

PR #1541, opened 2025-01-23
Ref: split settings into smaller screens

msalcala11:
Improves performance on older devices.

### DIFF
--- a/src/navigation/tabs/settings/SettingsRoot.tsx
+++ b/src/navigation/tabs/settings/SettingsRoot.tsx
@@ -1,8 +1,9 @@
 import {NativeStackScreenProps} from '@react-navigation/native-stack';
-import React, {ReactElement, useMemo, useRef} from 'react';
+import React, {useMemo, useRef} from 'react';
 import {useTranslation} from 'react-i18next';
-import {View, LayoutAnimation, ScrollView} from 'react-native';
+import {View} from 'react-native';
 import styled from 'styled-components/native';
+import {FlashList} from '@shopify/flash-list';
 import AngleRight from '../../../../assets/img/angle-right.svg';
 import Avatar from '../../../components/avatar/BitPayIdAvatar';
 import {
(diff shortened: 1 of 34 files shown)


src/navigation/tabs/settings/SettingsRoot.tsx
```diff
@@ -11,49 +12,42 @@ import {
...
-  flex: 1;
-`;
+export type SettingsHomeProps = NativeStackScreenProps<
+  SettingsGroupParamList,
+  SettingsScreens.SETTINGS_HOME
+>;
 
-export const Settings = styled.ScrollView`
-  padding: 10px ${ScreenGutter};
+export const SettingsContainer = styled(SafeAreaView)`
```
Gamboster, line 40: Any reason why we changed the ScrollView to a SafeAreaView? This is causing the listings to stop being scrollable. For example in the "Wallets & Keys" listing, contacts, external services history, languages ​​listing, etc.

src/navigation/tabs/settings/SettingsRoot.tsx
```diff
@@ -11,49 +12,42 @@ import {
...
+  flex: 1;
 `;
 
 export const SettingsComponent = styled.View`
```
Gamboster, line 44: Modifying this code seems to fix the scrolls

export const SettingsComponent = styled.ScrollView`
  /* flex: 1; */
  padding: 10px 0;
`;

Gamboster, line 44: Modifying this code seems to fix the scrolls

export const SettingsComponent = styled.ScrollView`
  /* flex: 1; */
  padding: 10px 0;
`;

src/navigation/tabs/settings/SettingsRoot.tsx
msalcala11: Good catch. Now fixed.

Gamboster (approved):
LGTM

PR #1544, opened 2025-01-27
Fix: set backdropColor instead white background

cmgustavo:


cmgustavo:
Closed in favor of https://github.com/bitpay/bitpay-app/pull/1560

PR #1572, opened 2025-02-14
Fix: add padding to "Slide to Send" button on Confirm Screen (Android)

cmgustavo:
It doesn't affect to iOS devices.


cmgustavo:
Closed in favor of https://github.com/bitpay/bitpay-app/pull/1592

PR #1594, opened 2025-02-20
Fix: add custom TouchableOpacity component to fix android modal pressability

msalcala11:
Uses TouchableOpacity from react-native on android (unless the component is inside of a navigation header), and uses TouchableOpacity from react-native-gesture-handler everywhere else (since the onPress handler for TouchableOpacity from react-native doesn't always fire on Mac desktop)

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -19,6 +19,16 @@ const plugins = [
       allowUndefined: true,
     },
   ],
+  [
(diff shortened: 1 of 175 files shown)


gabrielbazan7 (dismissed):
ACK. great job

gabrielbazan7 (approved):

PR #1598, opened 2025-02-21
Ref: new max limits for Moonpay-buy

Gamboster:
>Luke Kennedy (From MoonPay team): Maximum Buy Limit - Looks like the maximum buy is hardcoded to $20,000. Those who want to put large transactions through would be happy to know that we support transaction sizes up to $30,000; it could be worth somehow indicating this to users for the MoonPay option in your upstream UI.


JohnathanWhite (approved):

JohnathanWhite:
Thanks @Gamboster! I reached out to Connor regarding messaging

PR #1611, opened 2025-03-03
Fix: wallet connect pressables on android

msalcala11:
Currently, TouchableOpacity pressability inside of a sheet modal only works on android if it is imported from react-native, whereas TouchableOpacity inside of a header only works if it is imported from react-native-gesture-handler. This PR adds a new optional `touchableLibrary` param to the custom TouchableOpacity component that makes it possible to configure the import paths of TouchableOpacity components appropriately based on where they are placed to ensure onPress handlers fire on android.


gabrielbazan7 (approved):
ACK!

PR #1615, opened 2025-03-05
Fix: show ScanIcon on AddContacts screen

cmgustavo:
Current issue only happens on iOS.

Fix works for both (Android and iOS)

Simulator Screenshot - iPhone 16 Pro - 2025-03-05 at 10 24 19


gabrielbazan7 (approved):
ACK

PR #1658, opened 2025-03-31
Fix: TXP notifications checkbox workaround

Gamboster:


gabrielbazan7 (approved):
ACK. tested on Android ( device and simulator ), desktop and iOS ( simulator )

PR #1672, opened 2025-04-10
Perf: only encrypt sensitive wallet store fields to improve performance

msalcala11:
Moves `WALLET` store to `unencryptedStores`, and adds a new redux-persist transform that encrypts only sensitive `WALLET` fields before persisting the `WALLET` state. This greatly reduces encryption/decryption time, which boosts overall app performance, especially on older devices.


gabrielbazan7 (approved):
ACK !

PR #1675, opened 2025-04-15
Fix: get rates when app is installed for first time

cmgustavo:
* Show Exchange Rates at Home when install the app from scratch.


gabrielbazan7 (approved):
LGTM

PR #1688, opened 2025-04-22
Perf: only encrypt sensitive APP and SHOP store fields to boost performance

msalcala11:
Moves `APP` and `SHOP` stores to `unencryptedStores` and updates the `encryptSpecificFields` transform to only encrypt `APP.identity.priv` and sensitive `SHOP` store fields to reduce encryption/decryption time. Also ensures that gift card invoice objects are never persisted to the store as they are not used anywhere in the current codebase.

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -231,7 +231,7 @@ const getStore = async () => {
             ),
           );
         },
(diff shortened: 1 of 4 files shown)


cmgustavo (approved):
LGTM! 💪🏼

PR #1690, opened 2025-04-23
Fix: Amount and AmountModal UI

Gamboster:
This PR should fix cases where the "Continue" button appears below the UI, making it inaccessible for Android users.
It also addresses various cases where Archax banners are displayed, especially on smaller devices, preventing elements from collapsing.


cmgustavo (approved):
ACK! 👍🏼

PR #1691, opened 2025-04-23
Fix: init in amount page when fiat code is not found in rates

Gamboster:


src/components/amount/Amount.tsx
```diff
@@ -398,11 +398,24 @@ const Amount: React.FC<AmountProps> = ({
       !primaryIsFiat &&
       getRateByCurrencyName(allRates, currency.toLowerCase(), chain!)
     ) {
-      const fiatRate = getRateByCurrencyName(
+      const RateByCurrencyName = getRateByCurrencyName(
```
cmgustavo, line 401: This constant should not be capitalized

src/components/amount/Amount.tsx
```diff
@@ -421,7 +434,12 @@ const Amount: React.FC<AmountProps> = ({
   };
 
   useEffect(() => {
-    initRef.current();
+    try {
+      initRef.current();
+    } catch (err: any) {
+      logger.error('Error trying to init Amount view.');
+      logger.debug(err);
```
cmgustavo, line 441: Maybe this could be improved using:
```
const errStr = err instanceof Error ? err.message : JSON.stringify(err);
logger.error('[Amount] could not initialize view: ', errStr);
```

src/components/amount/Amount.tsx
Gamboster: done

src/components/amount/Amount.tsx
Gamboster: Done. Thanks!

cmgustavo (approved):
ACK!

PR #1700, opened 2025-05-05
Fix: chain extraction logic in ChainSearch. Ensure only selected chai…

cmgustavo:
…ns are included


gabrielbazan7 (approved):
ACK. Good catch

gabrielbazan7:
https://github.com/bitpay/bitpay-app/pull/1701

PR #1702, opened 2025-05-06
Fix: checks for price charts and fiatCurrency in rate cache

Gamboster:
This PR should also correct an inconsistency between the chart and the selected fiat currency.
Without this fix, if you change the fiat currency in settings and go to the charts before the cached time expires, it will display the current value correctly but the chart of the previously selected fiat currency.

iPhone_15_Pro


gabrielbazan7 (approved):
ACK

JohnathanWhite (approved):

PR #1743, opened 2025-06-16
[REF] Sol signing methods

gabrielbazan7:


src/components/modal/in-app-notification/InAppNotification.tsx
```diff
@@ -134,7 +134,7 @@ const InAppNotification: React.FC = () => {
                 {...{
                   width: 20,
                   height: 20,
-                  color: theme.dark ? White : Black,
```
JohnathanWhite, line 137: was this intentional?

JohnathanWhite (approved):

PR #1818, opened 2025-07-25
Fix: Archax banner on android. Also remove Exchange Rates from Home S…

cmgustavo:
…creen

### DIFF
--- a/src/navigation/tabs/TabContainer.tsx
+++ b/src/navigation/tabs/TabContainer.tsx
@@ -2,18 +2,19 @@ import React, {memo, useMemo, ReactNode} from 'react';
 import {Platform} from 'react-native';
 import styled from 'styled-components/native';
 import {useSafeAreaInsets} from 'react-native-safe-area-context';
+import {useAppSelector} from "../../utils/hooks";
 
 type PropsWithMoreParams<P = unknown> = P & {
   children: ReactNode;
-  removeMarginTop?: boolean;
 };
 
 const TabContainer: React.FC<PropsWithMoreParams> = ({children}) => {
   const insets = useSafeAreaInsets();
+  const showArchaxBanner = useAppSelector(({APP}) => APP.showArchaxBanner);
   const Container = useMemo(
     () => styled.View`
       flex: 1;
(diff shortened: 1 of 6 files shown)


src/navigation/tabs/settings/SettingsRoot.tsx
```diff
@@ -191,7 +191,7 @@ const SettingsHome: React.FC<SettingsHomeProps> = ({route, navigation}) => {
   );
 
   return (
-    <TabContainer>
+    <TabContainer removeMarginTop={showArchaxBanner ? true : false}>
```
JohnathanWhite: could be simplified to just removeMarginTop={showArchaxBanner}

src/navigation/tabs/TabContainer.tsx
```diff
@@ -8,12 +8,12 @@ type PropsWithMoreParams<P = unknown> = P & {
   removeMarginTop?: boolean;
 };
 
-const TabContainer: React.FC<PropsWithMoreParams> = ({children}) => {
+const TabContainer: React.FC<PropsWithMoreParams> = ({children, removeMarginTop}) => {
```
msalcala11: Rather than adding a new prop here, would it be simpler to place the following line at the top of the `TabContainer` component?:

`  const showArchaxBanner = useAppSelector(({APP}) => APP.showArchaxBanner);
`

We could then use the `showArchaxBanner` variable to conditionally show the top insets on android and then we wouldn't have to modify the 4 top-level tab files at all and can eliminate all instances of:
`<TabContainer removeMarginTop={showArchaxBanner ? true : false}>`

msalcala11 (changes requested):

src/navigation/tabs/TabContainer.tsx
cmgustavo: I agree! Thanks

src/navigation/tabs/settings/SettingsRoot.tsx
cmgustavo: Removed! Thanks

PR #1833, opened 2025-08-05
Feat: list all supported currencies and show if it is temporarily dis…

cmgustavo:
…abled


JohnathanWhite:
https://bitpayprod.atlassian.net/browse/RN-1870

PR #1842, opened 2025-08-11
Feat: list all supported currencies and show if it is temporarily disabled

cmgustavo:
* Fixed click on wallet selector
* Show all supported currencies by BitPay
* Label "Temporarily disabled"
* Label "Not supported"


gabrielbazan7 (dismissed):
ACK!

PR #1849, opened 2025-08-13
Feat: sardine SEPA payment method support

Gamboster:


JohnathanWhite (approved):

src/navigation/services/buy-crypto/constants/BuyCryptoConstants.tsx
```diff
@@ -137,7 +137,7 @@ export const PaymentMethodsAvailable: PaymentMethods = {
       banxa: true,
       moonpay: true,
```
Blend1221, line 138: True

PR #1880, opened 2025-09-03
Fix: avoid undefined value instead an empty Array

cmgustavo:
Steps to reproduce:

* Fresh install
* Import a key
* From Home Screen, go to "Send" button
* App crashes for an undefined value


Gamboster (approved):
LGTM

PR #1883, opened 2025-09-08
Fix: use ata toAddress in swap and sell for solana tokens

Gamboster:
This also includes a fix to show Solana tokens in Moonpay-sell


gabrielbazan7 (approved):
ACK

PR #1885, opened 2025-09-10
[FIX] Transfer SPL tokens to a wallet with no ATA created

gabrielbazan7:


src/store/wallet/effects/send/send.ts
```diff
@@ -1290,6 +1303,38 @@ export const publishAndSign =
         }
       }
 
+      if (ataOwnerAddress && txp.tokenAddress) {
```
Gamboster: can you add an isSvmChain check here to prevent possible errors?

PR #1886, opened 2025-09-11
Fix: create destination ataAddress if necessary in swap/sell

Gamboster:
This PR needs: https://github.com/bitpay/bitpay-app/pull/1885

### DIFF
--- a/src/navigation/services/swap-crypto/screens/ChangellyCheckout.tsx
+++ b/src/navigation/services/swap-crypto/screens/ChangellyCheckout.tsx
@@ -712,7 +712,7 @@ const ChangellyCheckout: React.FC = () => {
             key,
             wallet: fromWalletSelected,
             transport,
-            ataOwnerAddress: IsSVMChain(chain) ? ataOwnerAddress : undefined,
+            ataOwnerAddress,
           }),
         );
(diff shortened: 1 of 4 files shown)


src/navigation/services/swap-crypto/screens/ChangellyCheckout.tsx
```diff
@@ -693,19 +712,22 @@ const ChangellyCheckout: React.FC = () => {
...
       } else {
         dispatch(startOnGoingProcessModal('SENDING_PAYMENT'));
+        const {chain} = fromWalletSelected.credentials;
         await sleep(400);
         await dispatch(
           publishAndSign({
             txp: ctxp! as TransactionProposal,
             key,
             wallet: fromWalletSelected,
+            ataOwnerAddress: IsSVMChain(chain) ? ataOwnerAddress : undefined,
```
gabrielbazan7: You could remove that check and simply send the `ataOwnerAddress`. I have already added an `IsSVMChain(chain)` check inside publishAndSign.

Gamboster: done

PR #1926, opened 2025-10-03
[REF] braze - force token generation for iOS/Android

gabrielbazan7:


src/store/app/app.effects.ts
```diff
@@ -866,6 +867,8 @@ export const checkNotificationsPermissions = async (): Promise<boolean> => {
     normalized === RESULTS.GRANTED || normalized === RESULTS.LIMITED;
   if (granted) {
     Braze.requestPushPermission();
+    if (Platform.OS === 'ios')
```
cmgustavo: Missing token happens on iOS and Android. We shouldn't limit the same functionality to only iOS

src/store/bitpay-id/bitpay-id.effects.ts
```diff
@@ -119,10 +119,11 @@ export const startBitPayIdAnalyticsInit =
           agreedToMarketingCommunications,
         ),
       );
-      
       const systemEnabled = await checkNotificationsPermissions();
       if (systemEnabled && notificationsAccepted) {
-        dispatch(setNotifications(true));
+        Braze.setPushNotificationSubscriptionType(
```
cmgustavo: It's not necessary. `setNotifications(true)` already call to 
```
const value = accepted
      ? Braze.NotificationSubscriptionTypes.OPTED_IN
      : Braze.NotificationSubscriptionTypes.UNSUBSCRIBED;
```

src/store/app/app.effects.ts
gabrielbazan7: ok, I haven't checked the Android case yet. We could do that in a separate pull request.

cmgustavo: For now iOS should be enough

src/store/bitpay-id/bitpay-id.effects.ts
gabrielbazan7: This is necessary because setNotifications was subscribing wallets to BWS each time the app was refreshed, which I didn't want.

PR #2077, opened 2026-03-12
UI: Fix - startup navigation flicker on Android devices

cmgustavo:
Stabilize root navigation during app startup

### Changes
- Introduce `StartupGate` as the initial root route
- Defer startup navigation until `appWasInit` is ready
- Remove early splash hiding and hide it only after navigation leaves `StartupGate`
- Disable startup transition animation to avoid visible flicker
- Prevent duplicated `"Swiped Feature"` analytics event from `OnboardingStart`

### Result
- No incorrect onboarding/home flashes during app startup
- Cleaner splash-to-screen transition
- Cleaner onboarding analytics tracking

https://github.com/user-attachments/assets/dc991c67-6afe-4abd-acb1-2a51b88c33f6

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -2,6 +2,7 @@ import {
   NavigationContainer,
   NavigationState,
   NavigatorScreenParams,
+  useNavigation,
 } from '@react-navigation/native';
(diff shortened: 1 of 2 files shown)


gabrielbazan7 (approved):
ACK

PR #2089, opened 2026-03-18
Create/Import: Enhancement - remove duplicate work and introducing controlled concurrency

cmgustavo:
This PR improves app-side performance for `createKey` and `serverAssistedImport` by removing duplicate wallet creation and introducing controlled concurrency.

---
[RN-2647](https://bitpayprod.atlassian.net/browse/RN-2647)

### DIFF
--- a/src/store/wallet/effects/create/create.ts
+++ b/src/store/wallet/effects/create/create.ts
@@ -61,6 +61,35 @@ import {tokenManager} from '../../../../managers/TokenManager';
 import {logManager} from '../../../../managers/LogManager';
 import {Analytics} from '../../../analytics/analytics.effects';
 
+const runWithConcurrency = async <T, R>(
+  items: T[],
+  limit: number,
+  worker: (item: T, index: number) => Promise<R>,
+): Promise<R[]> => {
+  const results: R[] = new Array(items.length);
+  let nextIndex = 0;
+
+  const runners = Array.from(
+    {length: Math.min(limit, items.length)},
+    async () => {
+      while (true) {
+        const currentIndex = nextIndex++;
+        if (currentIndex >= items.length) {
+          return;
+        }
+        results[currentIndex] = await worker(items[currentIndex], currentIndex);
+      }
+    },
+  );
+
+  await Promise.all(runners);
+  return results;
+};
+
+const defer = (fn: () => void) => {
+  setTimeout(fn, 0);
+};
+
 export interface CreateOptions {
   network?: Network;
   account?: number;
@@ -121,6 +150,7 @@ export const startCreateKey =
         );
 
         const key = buildKeyObj({key: _key, wallets});
+
         if (context !== 'onboarding') {
           dispatch(Analytics.track('Created Key'));
         }
@@ -129,6 +159,7 @@ export const startCreateKey =
             key,
           }),
         );
+
         resolve(key);
       } catch (err) {
         reject(err);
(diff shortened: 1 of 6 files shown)


src/store/wallet/effects/create/create.ts
```diff
@@ -61,6 +61,39 @@ import {tokenManager} from '../../../../managers/TokenManager';
...
+      }
+    },
+  );
+
+  await Promise.all(runners);
+  return results;
+};
+
+const shouldHydrateAddressImmediately = (chain: string) => {
+  return ['btc', 'eth', 'matic'].includes(chain);
```
gabrielbazan7: This list feels arbitrary

src/store/wallet/effects/create/create.ts
gabrielbazan7: You’re creating addresses for all the base wallets here, but createWalletAddress is called again for those same wallets below.

src/store/wallet/effects/create/create.ts
```diff
@@ -491,6 +549,37 @@ export const createMultipleWallets =
...
+          };
+          dispatch(subscribeEmailNotifications(wallet, prefs));
+        }
+
+        if (
+          !wallet.receiveAddress &&
+          shouldHydrateAddressImmediately(wallet.credentials.chain) === false &&
+          !wallet.credentials.token
+        ) {
+          dispatch(createWalletAddress({wallet, newAddress: true}));
```
gabrielbazan7: I’m not sure it’s worth deferring wallet address creation. Although it could save a few seconds during create, the address is needed to build the accounts list, and this might introduce potential errors.

src/store/wallet/effects/create/create.ts
```diff
@@ -451,38 +487,60 @@ export const createMultipleWallets =
             }
           }
         }
+        const chainTokens = tokens.filter(token => token.chain === coin.chain);
+
+        return {coin, wallet, chainTokens};
       } catch (err) {
-        const errMsg = err instanceof Error ? err.message : JSON.stringify(err);
-        logManager.debug(`Error creating wallet - continue anyway: ${errMsg}`);
+        return null;
```
gabrielbazan7, line 480: I would re-add the logs here

src/store/wallet/effects/create/create.ts
gabrielbazan7: same with createTokenWallet

src/store/wallet/effects/create/create.ts
cmgustavo: Good point. The intent was to keep critical-path address hydration limited to a small priority subset so `createKey` completes faster, not to special-case those chains permanently. Just moved it to a constant and added a short comment.

src/store/wallet/effects/create/create.ts
cmgustavo: Good catch! just removed duplicated functions. The complete flow is:

1. create base wallets
2. hydrate priority addresses once
3. create token wallets once
4. defer subscriptions + non-priority address hydration

src/store/wallet/effects/create/create.ts
cmgustavo: Agreed. I reverted the deferred address hydration change so base wallets still create their address during the main flow. This keeps the wallet/address contract unchanged while preserving the concurrency improvement for wallet creation.

src/store/wallet/effects/create/create.ts
cmgustavo: +1

src/store/wallet/effects/create/create.ts
cmgustavo: Removed

gabrielbazan7 (dismissed):
ACK. Great work !

JohnathanWhite (approved):

PR #2108, opened 2026-03-31
AppsFlyer: Fix - Prevent unhandled exceptions and SDK calls when not initialized

cmgustavo:
RN-2412
RN-2416

## Changes
- Remove public configureResolvedDeepLinks() from the wrapper API
- Configure wrapped deep link hosts internally during init()
- Keep AppsFlyer state handling in the wrapper to **avoid SDK calls before successful initialization**

## Notes

Wrapped deep link configuration is startup-time SDK setup, not a general runtime action. Keeping it inside init() simplifies call sites and avoids calling deep link configuration separately or when the SDK is not ready.

---
[RN-2412](https://bitpayprod.atlassian.net/browse/RN-2412)

### DIFF
--- a/src/constants/config.ts
+++ b/src/constants/config.ts
@@ -30,6 +30,12 @@ export const STATIC_CONTENT_CARDS_ENABLED = parseEnvBoolean(
 );
 export const APP_ANALYTICS_ENABLED = !__DEV__;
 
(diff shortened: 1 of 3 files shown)


gabrielbazan7 (approved):
ACK! great job !

PR #2110, opened 2026-03-31
Mixpanel: Enhancement - Prevent multiple initialization and upgrade plugin

cmgustavo:
### Changes

* Added stateful initialization logic similar to AppsFlyer
* Ensured identify, track, and reset are only executed when SDK is ready
* Normalized track() to always pass an object (properties ?? {})
* Upgrade to latest `mixpanel-react-native` plugin v3.3.0


gabrielbazan7 (approved):
code ACK!

PR #2112, opened 2026-04-01
Analytics: Fix - Prevent multiple SDK initializations

cmgustavo:
RN-2412

### changes

- Multiple calls to `initialize()` will reuse the same promise
- SDK initialization is guaranteed to run only once
- Subsequent calls will await the ongoing initialization instead of re-triggering it
- Modified getTrackingAuthorizedByUser() to use async/await and properly waits for the tracking permission
- Removed redundant Analytics.initialize() calls from UI-level handlers to avoid unintended re-initialization attempts


src/store/analytics/analytics.effects.ts
```diff
@@ -76,66 +82,82 @@ export const Analytics = (() => {
...
+            await dispatch(getTrackingAuthorizedByUser());
+          }
 
-      logManager.info('Successfully initialized analytics.');
+          _isInitialized = true;
+          flushQueue();
 
-      return;
+          logManager.info('Successfully initialized analytics.');
+        } finally {
```
gabrielbazan7, line 101: let's add a catch here: 

```
catch (err) {
          const errMsg =
            err instanceof Error ? err.message : JSON.stringify(err);
          logManager.error(`[Analytics] init failed: ${errMsg}`);
        }
```

src/store/analytics/analytics.effects.ts
```diff
@@ -6,46 +6,52 @@ import {MixpanelWrapper} from '../../lib/Mixpanel';
...
   let _preInitQueue: Array<() => void> = [];
   let _isInitialized = false;
+  let _isInitializing = false;
+  let _initPromise: Promise<void> | null = null;
   let _isMergingUser = false;
 
+  const flushQueue = () => {
+    const queue = _preInitQueue;
+    _preInitQueue = [];
+    queue.forEach(cb => cb());
```
gabrielbazan7, line 51: Let's add a catch here for the queue: 

```
  queue.forEach(cb => {
      try {
        cb();
      } catch (err) {
        const errMsg = err instanceof Error ? err.message : JSON.stringify(err);
        logManager.error(`[Analytics] Error executing queued event: ${errMsg}`);
      }
    });
```

src/store/analytics/analytics.effects.ts
```diff
@@ -6,46 +6,52 @@ import {MixpanelWrapper} from '../../lib/Mixpanel';
...
+      'An error occurred while requesting tracking permission',
+      errMsg,
+    );
+  }
+};
 
 export const Analytics = (() => {
   let _preInitQueue: Array<() => void> = [];
   let _isInitialized = false;
+  let _isInitializing = false;
```
gabrielbazan7: `_isInitializing` is not used. Let's remove it

src/store/analytics/analytics.effects.ts
cmgustavo: Not necessary since each individual SDK init has already try/catch in lines 98, 111, and 139.

src/store/analytics/analytics.effects.ts
cmgustavo: It is only useful if it’s written to handle async rejections, not just sync exceptions

src/store/analytics/analytics.effects.ts
cmgustavo: Removed! if we need it in a future, we can add it.

gabrielbazan7 (approved):

PR #2124, opened 2026-04-07
Send: Feature - Nonce assignment on signing for EVM and XRP

leolambo:
### Description
[RN-2649](https://bitpayprod.atlassian.net/browse/RN-2649)
replaces https://github.com/bitpay/bitpay-app/pull/2123

EVM and XRP transaction proposals can now have their nonce deferred to signing time. When the app sees a txp with `deferNonce` set or a null nonce on a nonce-based chain, it calls `prepareTx` to get a fresh nonce from BWS right before signing. This prevents stale nonces on proposals that sit pending.

Also widens the multi-proposal signing filters so XRP txps go through the sequential path alongside EVM, since they both depend on ordered nonce assignment.

**Depends on:**
- BWS/BWC: https://github.com/bitpay/bitcore/pull/4124 (prepareTx endpoint + deferNonce support)
- Server: internal MR 11390 (deferNonce runtime setting)

### Changelog

* Added `IsNonceChain` helper covering EVM + XRP chains
* `publishAndSign` calls `prepareTx` before signing when nonce is missing on a nonce chain
* Multi-proposal filters use `IsNonceChain` instead of just `IsEVMChain` so XRP txps are signed sequentially
* `deferNonce` field added to `TransactionProposal` model

### Testing Notes
With `deferNonce` **disabled** in runtime settings, behavior should be identical to before — nonces are assigned at creation time.

With `deferNonce` **enabled**:
1. Create an EVM or XRP txp — should have null nonce and `deferNonce: true`
2. Sign the txp — app should log `prepareTx: BWS assigned nonce X to txp Y` before signing
3. Bulk sign multiple EVM/XRP txps — they should be processed sequentially with incrementing nonces

UTXO chains (BTC, BCH, etc.) are unaffected.

### DIFF
--- a/src/store/wallet/effects/send/send.ts
+++ b/src/store/wallet/effects/send/send.ts
@@ -91,6 +91,7 @@ import {
   IsSVMChain,
   IsUtxoChain,
   IsEVMChain,
+  IsNonceChain,
 } from '../../utils/currency';
 import {CommonActions, NavigationProp} from '@react-navigation/native';
 import {BwcProvider} from '../../../../lib/bwc';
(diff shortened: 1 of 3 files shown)


gabrielbazan7 (approved):
Code ACK !

PR #2140, opened 2026-04-16
Braze: Enhancement - Refactor to ensure single SDK initialization

cmgustavo:


src/lib/Braze/index.ts
```diff
@@ -148,78 +149,151 @@ export type BrazeUserAttributes = {
...
+  merge(userToMerge: string, userToKeep: string) {
+    return mergeUsers(userToMerge, userToKeep);
+  }
+
+  delete(eid: string) {
+    return deleteUser(eid);
+  }
+
+  async screen(name: string, properties: Record<string, any> = {}) {
+    if (!(await this.ensureReady())) {
```
gabrielbazan7, line 260: let's use `if (!this.isReady()) {` here

src/lib/Braze/index.ts
```diff
@@ -148,78 +149,151 @@ export type BrazeUserAttributes = {
...
+      return;
+    }
+
+    const screenName = `Viewed ${name} Screen`;
+    Braze.logCustomEvent(screenName, properties);
+  }
+
+  async track(eventName: string, properties: Record<string, any> = {}) {
+    if (!(await this.ensureReady())) {
```
gabrielbazan7, line 269: let's use `if (!this.isReady()) {` here

src/lib/Braze/index.ts
cmgustavo: Differences between `isReady` is that `ensureReady` tries to initialize SDK with `initPromise`. It tries to re-initialize if there's error during the startup (using screen/tracker events)

src/lib/Braze/index.ts
cmgustavo: Same as above

JohnathanWhite (approved):

PR #2145, opened 2026-04-17
XCode: Fix - Patch fmt consteval for Xcode 26/Clang compatibility

cmgustavo:
Screenshot 2026-04-17 at 5 35 47 PM

### Summary

Xcode 26 ships with a stricter version of Clang that rejects `consteval` in certain contexts inside the fmt library (v11.0.2). Specifically, `fmt` defines a macro called `FMT_CONSTEVAL` that resolves to the C++ `consteval` keyword when the compiler supports it — but Clang 26 raises an error when that keyword is used inside a lambda that returns a local struct, which is exactly what `fmt's FMT_STRING` macro does internally.

### Fix

Added a `post_install` patch in the `Podfile` that rewrites `#define FMT_CONSTEVAL consteval` to `#define FMT_CONSTEVAL constexpr` in `fmt/base.h` after every pod install

---
[RN-2691](https://bitpayprod.atlassian.net/browse/RN-2691)

### DIFF
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -79,6 +79,12 @@ target 'BitPayApp' do
           config.build_settings['OTHER_SWIFT_FLAGS'] << '-no-verify-emitted-module-interface'
         end
       end
+      if target.name == 'fmt' || target.name == 'RCT-Folly'
+        target.build_configurations.each do |config|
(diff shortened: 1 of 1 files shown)


gabrielbazan7 (approved):
ACK

PR #2148, opened 2026-04-22
Fix: pre-select wallet from assets details correctly

Gamboster:
Ticket reference: https://bitpayprod.atlassian.net/browse/RN-2695

This fix should resolve the issue where clicking on swap displays an "Insufficient Funds" message in the Global Selector.

---
[RN-2695](https://bitpayprod.atlassian.net/browse/RN-2695)


cmgustavo (approved):
ACK!

PR #2151, opened 2026-04-23
Braze: Fix - Prevent remove oldEid before merge data to newEid

cmgustavo:
### Problem
Anonymous Braze tracking data was not migrating to the authenticated user profile on login.

### Root cause
`Braze.changeUser(newEid)` and the server-side merge were called in the wrong order, and buffered SDK events were never flushed before switching users.

### Fix
* Added `Braze.requestImmediateDataFlush()` before `Braze.changeUser()` in `BrazeWrapper.identify` to flush buffered anonymous events to Braze servers first.
* Reordered the login flow so `identify` (→ `changeUser`) runs before the server-side merge API call.
* Removed programmatic delete of the anonymous profile — it races against Braze's async merge and can cancel it. The empty shell is harmless and handled by Braze's User Archival policy.

---
[RN-2699](https://bitpayprod.atlassian.net/browse/RN-2699)

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -145,7 +145,6 @@ import {
   successAddWallet,
   successGetReceiveAddress,
 } from './store/wallet/wallet.actions';
-import {BrazeWrapper} from './lib/Braze';
 import {selectSettingsNotificationState} from './store/app/app.selectors';
 import {HeaderShownContext} from '@react-navigation/elements';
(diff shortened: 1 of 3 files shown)


gabrielbazan7 (approved):
ACK

cmgustavo:
Closed in favor of https://github.com/bitpay/bitpay-app/pull/2152

PR #2152, opened 2026-04-23
Braze: Fix - No remove OldEid and subscribe new users to email notifications

cmgustavo:
### Commit f305ce4347e7aa99009934c602a6945c1744cba6

New users could end up with `UNSUBSCRIBED` email status in Braze because `agreedToMarketingCommunications` was propagated through the pairing chain and used to gate the setEmailNotifications call. 
Now **setEmailNotificationsAccepted(true)** is dispatched immediately after account creation, and the `agreedToMarketingCommunications` parameter is removed from the chain. 

The OPTED_IN vs SUBSCRIBED distinction is determined solely by the BitPay user's optInEmailMarketing + verified flags.

[RN-2700](https://bitpayprod.atlassian.net/browse/RN-2700)

---

### Commit a8235395bb8d9b09a8d8a5b3fe35243e574d9fc2

Anonymous Braze tracking data was not migrating to the authenticated user profile on login.

#### Root cause

Braze.changeUser(newEid) and the server-side merge were called in the wrong order, and buffered SDK events were never flushed before switching users.

#### Fix

Added Braze.requestImmediateDataFlush() before Braze.changeUser() in BrazeWrapper.identify to flush buffered anonymous events to Braze servers first.
Reordered the login flow so identify (→ changeUser) runs before the server-side merge API call.
Removed programmatic delete of the anonymous profile — it races against Braze's async merge and can cancel it. The empty shell is harmless and handled by Braze's User Archival policy.

[RN-2699](https://bitpayprod.atlassian.net/browse/RN-2699)

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -145,7 +145,6 @@ import {
   successAddWallet,
   successGetReceiveAddress,
 } from './store/wallet/wallet.actions';
-import {BrazeWrapper} from './lib/Braze';
 import {selectSettingsNotificationState} from './store/app/app.selectors';
 import {HeaderShownContext} from '@react-navigation/elements';
 import PaymentSent from './navigation/wallet/components/PaymentSent';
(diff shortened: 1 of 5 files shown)


gabrielbazan7 (dismissed):
ACK

gabrielbazan7 (approved):
ACK

PR #2154, opened 2026-04-28
Import: Enhancement - deferred import

gabrielbazan7:


src/store/app/app.reducer.ts
```diff
@@ -53,6 +53,8 @@ export const appReduxPersistBlackList: Array<keyof AppState> = [
   'tokensDataLoaded',
   'isImportLedgerModalVisible',
   'showArchaxBanner',
+  'importBannerMessage',
+  'importIsFirstKey',
```
JohnathanWhite, line 57: **pendingImport and importProgress are persisted, so a killed import wedges the UI permanently**

Only `importBannerMessage` and `importIsFirstKey` were added to `appReduxPersistBlackList`; the two other new fields, `pendingImport` and `importProgress`, are left persisted. Because the deferred import now continues in a detached IIFE after the user has navigated away, an app kill or crash mid-import persists `pendingImport: true` with a partial `importProgress`, and nothing on the next launch resets them — `setPendingImport(false)` is only dispatched from that IIFE's success and catch paths, which are gone. On rehydrate the user gets the import skeleton card in the Home carousel and a stuck progress bar, plus `disabled={pendingImport}` on the create/import controls in `Crypto.tsx` and `WalletsAndKeys.tsx`, so they cannot add or import any key at all. `importIsFirstKey` resets to `false` while `pendingImport` stays `true`, which makes the state internally inconsistent as well. These four fields look like they should all be blacklisted together.

JohnathanWhite (commented):

PR #2155, opened 2026-04-28
Braze: Chore - Upgrade to latest plugin version

cmgustavo:
Upgrades the Braze React Native SDK to v20.1.0 and migrates to its new delayed-initialization pattern, where the API key and endpoint are no longer hardcoded in native config files but instead passed from the JS layer at runtime.

https://github.com/braze-inc/braze-react-native-sdk/blob/master/CHANGELOG.md#2010

### DIFF
--- a/src/lib/Braze/index.ts
+++ b/src/lib/Braze/index.ts
@@ -1,6 +1,12 @@
 import Braze from '@braze/react-native-sdk';
 import axios from 'axios';
-import {BRAZE_MERGE_AND_DELETE_API_KEY, BRAZE_REST_API_ENDPOINT} from '@env';
+import {
+  BRAZE_API_ENDPOINT,
+  BRAZE_API_KEY_ANDROID,
+  BRAZE_API_KEY_IOS,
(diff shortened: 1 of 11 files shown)


src/lib/Braze/index.ts
```diff
@@ -1,6 +1,12 @@
 import Braze from '@braze/react-native-sdk';
 import axios from 'axios';
-import {BRAZE_MERGE_AND_DELETE_API_KEY, BRAZE_REST_API_ENDPOINT} from '@env';
+import {
+  BRAZE_API_ENDPOINT,
```
gabrielbazan7, line 4: This is missing from the declarations.d.ts file

gabrielbazan7 (dismissed):

src/lib/Braze/index.ts
cmgustavo: Fixed! thanks

gabrielbazan7 (approved):
ACK!

JohnathanWhite:
Synced to Jira: https://bitpayprod.atlassian.net/browse/RN-2805

PR #2157, opened 2026-05-04
iOS uitests added

vinotestlio:
- Added XCUITest smoke tests
- Implemented Page Object Model (POM) structure with relevant pages
- Integrated Allure reporting for test results
- Added GitHub Actions workflow to execute tests on macOS runner


android/app/src/main/res/xml/network_security_config.xml
```diff
@@ -1,7 +1,7 @@
 <network-security-config>
   <!-- Set cleartextTrafficPermitted = true for development -->
-  <base-config cleartextTrafficPermitted="false" />
-  <domain-config cleartextTrafficPermitted="false">
+  <base-config cleartextTrafficPermitted="true" />
```
JohnathanWhite, line 3: **Blocking.** This is `scripts/dev-overrides.js:12-15` output — it does exactly this `replaceAll` on `cleartextTrafficPermitted`. The file is referenced from `src/main`, so `false` -> `true` on `base-config` disables Android's HTTPS enforcement app-wide in **release builds** of a cryptocurrency wallet.

Please `git checkout origin/develop -- android/app/src/main/res/xml/network_security_config.xml` and check `git status` after any `yarn set:dev`.

android/app/src/main/java/com/bitpay/wallet/MainActivity.kt
```diff
@@ -48,10 +48,15 @@ class MainActivity : ReactActivity() {
         (application as MainApplication).addActivityToStack(this.javaClass)
 
         window.apply {
-            setFlags(
-                WindowManager.LayoutParams.FLAG_SECURE,
-                WindowManager.LayoutParams.FLAG_SECURE
-            )
+            if (!BuildConfig.DEBUG) {
```
JohnathanWhite, line 51: Non-blocking — this is correctly scoped to debug and release is unaffected, so I'm fine with it in principle.

It does mean debug builds can screenshot the recovery-phrase screen, which is a deliberate loosening of a security control arriving inside a test PR. Could you get an explicit nod from someone on the app side rather than having it ride along here?

android/build.gradle
```diff
@@ -50,5 +50,49 @@ allprojects {
         maven { url 'https://www.jitpack.io' }
         maven { url "https://appboy.github.io/appboy-android-sdk/sdk" }
         maven { url "https://dl.appsflyer.com" }
+        maven { url "https://maven.sumsub.com/repository/maven-public/" }
     }
 }
+
+gradle.afterProject { proj ->
```
JohnathanWhite: **Blocking.** The `minSdkVersion` guard is fine (`:app` is 30, so it never fires). The `packagingOptions` block below has no guard and applies to every subproject **including `:app`**, so it reaches the release APK:

- `pickFirst` on `libhermes.so` / `libjsi.so` / `libreactnative.so` / `libc++_shared.so` / `libfbjni.so` silences the duplicate-native-library error. That error is the guard against an ABI mismatch between a vendored module's copy and `react-android`'s — silencing it globally turns a build failure into a runtime crash that only appears on user devices.
- `excludes += ['META-INF/LICENSE', 'META-INF/NOTICE', ...]` strips third-party licence and notice files from the shipping APK. Apache-2.0 requires NOTICE redistribution, so that's a compliance issue.
- `excludes += 'META-INF/*.kotlin_module'` can break Kotlin top-level function resolution and reflection at runtime.
- `compileSdkVersion 36` is forced onto any subproject declaring < 34, compiling third-party modules against an SDK they never declared.

If this is only needed for `assembleDebugAndroidTest`, please scope it — `if (proj.name == 'app')` plus a variant filter, or move it into the androidTest configuration. This file is also one of the 14 merge conflicts against `develop`.

android/app/src/androidTest/java/com/bitpay/wallet/tests/Test592ImportWalletRecoveryPhrase.kt
```diff
@@ -0,0 +1,102 @@
...
+        allureStep("Click Add Your Crypto on Home page") {
+            homePage.clickAddYourCrypto()
+        }
+
+        allureStep("Click Import Key option") {
+            addYourCryptoOptionPage.clickImportKey()
+        }
+
+        allureStep("Enter recovery phrase and import wallet") {
+            importWalletPage.enterRecoveryPhrase("hobby short divert lady spare quit act settle body town license alone")
```
JohnathanWhite, line 81: **Blocking.** This is a valid mainnet BIP39 mnemonic committed in plaintext to a public repo — here and in 4 other Android tests and 1 iOS test.

I checked: the BIP44/49/84 BTC addresses and the first three `m/44'/60'` ETH addresses are all empty today, so nothing is at immediate risk. But the wallet is permanently burned, and `Test596BuyBTC` step 9 says *"tap My Key with funds available"* — so the suite expects it to be funded, and once it is, anyone who reads this PR can sweep it.

It also leaves two other ways:
- the `allureScreenshot("Recovery phrase entered")` on the next line captures it on screen, and that image is uploaded to Testlio and as a CI artifact;
- `android-espresso.yml` uploads the full `adb logcat` as the `device-logcat` artifact, and that workflow's own comment says RN dev logging serialises "every dispatched action" and "a recovery-phrase import fires hundreds of actions".

Please treat this phrase as compromised, generate a fresh fixture, and inject it from a GitHub secret. A secret in git history can't be removed by a later commit.

android/app/src/androidTest/java/com/bitpay/wallet/utils/RetryRule.kt
```diff
@@ -0,0 +1,29 @@
+package com.bitpay.wallet.utils
+
+import android.util.Log
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+
+class RetryRule(private val retryCount: Int = 1) : TestRule {
```
JohnathanWhite: `repeat(retryCount)` with `retryCount = 1` means one attempt — so `RetryRule` never actually retries.

`BaseTest.kt:37` documents "Defaults to 3" but `BaseTest.kt:50` is `retryCount: Int = 1`, so the rule, the `@Rule(order = ...)` ordering comment, and the screenshot-on-final-failure design are all inert as configured. Either make the default match the doc or drop the machinery — right now it reads as working retry logic that isn't.

android/app/src/androidTest/java/com/bitpay/wallet/pages/OnboardingPage.kt
```diff
@@ -0,0 +1,400 @@
...
+        } catch (e: Throwable) {
+            false
+        }
+    }
+    fun clickBackupRecoveryPhrase() {
+        WaitUtils.waitForView(backupRecoveryPhraseElement)
+        onView(backupRecoveryPhraseElement).perform(click())
+    }
+
+    fun verifyIUnderstandCheckbox1Displayed(): Boolean {
```
JohnathanWhite, line 194: This returns `true` when the checkbox is **absent** — and `clickIUnderstandCheckbox1/2/3` and `clickAgreeAndContinue` below all early-return in the same state, before their assertions.

So if a regression makes the app skip the mandatory terms-consent screen entirely, every one of those steps reports as passed and the suite goes green on exactly the failure it's there to catch.

The recovery behaviour itself is reasonable — it just needs to live in a separately named helper (`skipTermsIfAlreadyOnboarded()`) that can't satisfy an assertion. Also note the 420,000 ms (7 min) timeout on the fallback.

ios/BitPayAppUITests/tests/Test596_BasicBuyBTC_.swift
```diff
@@ -0,0 +1,76 @@
...
+    AllureXCTestSupport.step("(8) On the Select Crypto screen, tap Bitcoin (BTC). ") {
+      selectCurrencyPage.tapBitcoinCurrency()
+    }
+    
+    AllureXCTestSupport.step("(9) On the Select Key to Deposit to screen, tap My Key with funds available.") {
+      XCTAssertTrue(selectKeyToDepositPage.isSelectKeyToDepositToDisplayed(), "Select Key to Deposit to dialog not displayed")
+      selectKeyToDepositPage.tapMyKeyWallet()
+    }
+    
+    AllureXCTestSupport.step("(10) Wait until best-offer search completes and at least one offer is displayed.") {
```
JohnathanWhite: Steps 10, 11 and 12 have **empty bodies**. The test's own description is "Basic Buy BTC (Best Offer -> Continue With Opens Browser)" and none of that is exercised or asserted — the test passes whether or not offers ever load.

Either implement the three steps or drop them and rename the test to the boundary it actually reaches, so the Testlio report doesn't overstate coverage.

ios/BitPayAppUITests/tests/Test593_BTCConfirmPaymentScreen.swift
```diff
@@ -0,0 +1,82 @@
...
+        "Select Currency page not displayed"
+      )
+    }
+    
+    AllureXCTestSupport.step("(6) Tap Bitcoin.") {
+      selectCurrencyPage.tapBitcoinCurrency()
+    }
+    
+    AllureXCTestSupport.step("(7) Wait until the Send To screen is displayed.") {
+      selectCurrencyPage.isSendToTitleDisplayed()
```
JohnathanWhite: Result discarded — this step is labelled "(7) Wait until the Send To screen is displayed" but asserts nothing.

```suggestion
      XCTAssertTrue(
        selectCurrencyPage.isSendToTitleDisplayed(),
        "Send To screen not displayed"
      )
```

ios/BitPayAppUITests/flows/AppFlows.swift
```diff
@@ -0,0 +1,89 @@
...
+      if homeTab.exists && homeTab.isHittable {
+        homeTab.tap()
+      } else if backControl.exists && backControl.isHittable {
+        backControl.tap()
+      } else if app.navigationBars.buttons.firstMatch.exists
+        && app.navigationBars.buttons.firstMatch.isHittable {
+        app.navigationBars.buttons.firstMatch.tap()
+      } else {
+        
+        app.coordinate(withNormalizedOffset: CGVector(dx: 0.12, dy: 0.11)).tap()
```
JohnathanWhite, line 78: When none of the known controls is found, this falls through to tapping absolute screen coordinate `(0.12, 0.11)` and dragging from the left edge — up to 15 times, with no knowledge of what screen it's on.

`goToHomeScreen` runs in `setUp()` for the send, swap, sell and buy tests, so these blind taps fire immediately before the flows that move money. (The element-relative `element.coordinate(...)` calls in the page objects are fine — this is the only absolute one.)

Two requests: drop the blind-tap fallback in favour of an explicit failure, and make the function fail rather than return silently when it runs out of attempts. Right now the test proceeds from an unknown screen and the eventual failure points at the wrong step.

scripts/run-tests.sh
```diff
@@ -0,0 +1,34 @@
...
+  EXIT=$?
+  cd ..
+
+  [ $EXIT -ne 0 ] && OVERALL_EXIT=$EXIT
+
+  # Pull allure results immediately after each test
+  # so a timeout or crash on a later test doesn't lose earlier results
+  echo "=== Pulling allure results after $TEST_CLASS ==="
+  mkdir -p /home/runner/work/_temp/allure-results
+  adb pull /sdcard/googletest/test_outputfiles/allure-results/. /home/runner/work/_temp/allure-results/ || true
```
JohnathanWhite, line 31: Two things here:

1. `/home/runner/work/_temp/...` is hardcoded, so this script can't run outside a GitHub runner despite living in `scripts/`. Please take the output directory from an env var with a local default.
2. Combined with the `|| true` here, the `cp ... || true` in the workflow, and the "Check Allure results" step that emits `::warning::` and exits 0, total loss of results is a **green** step that then uploads an empty archive to Testlio. That last check should fail the job when no result JSON exists.

.github/workflows/android-espresso.yml
```diff
@@ -0,0 +1,272 @@
+name: Android Espresso E2E
+
+on:
+  workflow_run:
+    workflows: ["iOS XCUITest"]
```
JohnathanWhite, line 5: **Blocking-ish.** Three problems in this trigger:

1. `ios-xcuitest.yml` and `ios-xcuitest-sandbox.yml` are **both** `name: iOS XCUITest`, so this matches either. A manual sandbox run kicks off this job using the **production** Testlio token and project. (Both Android files likewise share `name: Android Espresso E2E`.)
2. `types: [completed]` with no `github.event.workflow_run.conclusion == 'success'` guard — the Android suite runs after a failed or cancelled iOS run.
3. Under `workflow_run`, the bare `actions/checkout@v4` below checks out the **default branch** (`master`), not the ref the triggering run tested. Relatedly, `schedule:` also only fires on the default branch, so the nightly won't run at all until this reaches `master`.

Please give the sandbox workflows distinct names, add the `conclusion == 'success'` guard, and check out `github.event.workflow_run.head_sha` explicitly.

.github/workflows/ios-xcuitest-sandbox.yml
```diff
@@ -0,0 +1,283 @@
...
+        if: always()
+        run: npm install -g @testlio/cli
+
+      - name: Upload Results to Testlio
+        if: always()
+        env:
+          TESTLIO_DEVICE_ID: 363a5ce7-6cef-436c-8531-6c07195c8fe8
+        run: |
+          if [ "${{ inputs.environment }}" = "Sandbox" ]; then
+            export RUN_API_TOKEN="${{ secrets.TESTLIO_RUN_TOKEN }}"
```
JohnathanWhite: Two notes:

1. Interpolating `${{ secrets.* }}` into the body of a `run:` script splices the value into the script text. Please pass them via `env:` instead and reference `$RUN_API_TOKEN`.
2. The `environment` input only picks which **Testlio workspace** receives the report — the app build is identical (`cp .env.template .env.development` + `yarn set:dev`), and `src/constants/config.ts:41` hardcodes `APP_NETWORK = Network.mainnet` with no env override. So "Sandbox" still runs on **mainnet against `https://bitpay.com` and `https://bws.bitpay.com`** and hits live Changelly/Moonpay endpoints. Either rename it so it doesn't imply isolation, or wire the input through to the app config.

ios/BitPayAppUITests/tests/TestPlan.xctestplan
```diff
@@ -0,0 +1,28 @@
...
+  "testTargets" : [
+    {
+      "target" : {
+        "containerPath" : "container:BitPayApp.xcodeproj",
+        "identifier" : "CDF085062FACD8A500CA4C1B",
+        "name" : "BitPayAppUITests"
+      }
+    }
+  ],
+  "version" : 1
```
JohnathanWhite, line 27: This test plan is never used: the scheme has no `TestPlanReference`/`<TestPlans>` and neither workflow passes `-testPlan`, so `testTimeoutsEnabled` and the attachment lifetimes here do nothing.

It's also wired into **Copy Bundle Resources** (`project.pbxproj:423`), so it ships inside the `.xctest` bundle as a resource rather than configuring anything. Either reference it from the scheme and pass `-testPlan` in CI, or drop it — and remove it from the Resources phase either way.

ios/BitPayAppUITests/pages/VerifyYourPhrasePage.swift
```diff
@@ -0,0 +1,100 @@
...
+        text.range(of: "^[a-z]+$", options: .regularExpression) != nil
+      return isWord ? text : nil
+    }
+    .prefix(12)
+    .map { $0 }
+  }
+
+  func extractNumbers(from text: String) -> [String] {
+    do {
+      let regex = try NSRegularExpression(pattern: "\\d+")
```
JohnathanWhite, line 58: This page object and `NewRecoveryPhrasePage.swift` aren't referenced by any test or flow, so this is currently dead code — happy for it to be dropped from the PR.

If it's kept: `VerifyPhrase.tsx` renders each index twice ("2nd (2)"), so `\d+` yields `[2,2,11,11,9,9]` for three fields, and the fourth loop iteration indexes past the end of `textFields`. Capture only the parenthesised numbers and bounds-check both subscripts. There's also a force-unwrapped `Range(...)!` just below and a hardcoded `.prefix(12)`.

JohnathanWhite (changes requested):
The branch merged `master` while targeting `develop`, so GitHub's "Files changed" tab is computed against a stale merge base and **does not show all of the production changes** — `src/navigation/wallet/screens/send/SendTo.tsx` is missing from it entirely. Diff against the real merge base (`git diff $(git merge-base upstream/develop <branch>) <branch>`) to see them: 66 files, not 205.

---

## Blocking

### 1. `SendTo.tsx` — the new email regex is a catastrophic-backtracking hang in the Send flow

`src/navigation/wallet/screens/send/SendTo.tsx:157`

```diff
-  return /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(text);
+  const reg = /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w\w+)+$/;
+  return reg.test(text);
```

The replacement has nested quantifiers (`(\w+)*` twice). Measured on Node 22:

| input | new regex | old regex |
|---|---|---|
| `"a"*18 + "@" + "a"*18 + "!"` | **77,724 ms** | < 0.01 ms |
| `"a"*20 + "@" + "a"*20 + "!"` | did not finish in 2 min | < 0.01 ms |

It's reachable on the hot path. `isEmailAddress` is called from `validateAndNavigateToConfirm`, which is wired to:

- `SendTo.tsx:629` — `SearchInput` `onChangeText`, i.e. **every keystroke** in "Search contact or enter address"
- `SendTo.tsx:647` — `onScanComplete`, i.e. **whatever a scanned QR code contains**

So a signed-in user typing or pasting any `<word chars>@<word chars><non-word>` string into the Send recipient field freezes the JS thread, and a hostile QR code does it without the user typing anything. The old pattern has no nested quantifier and cannot backtrack this way.

It's also a behaviour regression independent of the hang — measured, `user+tag@example.com` returns `true` under the old regex and `false` under the new one, so plus-addressed recipients stop being recognised.

I can't find anything in this PR that needs this change. **Please revert the hunk.** Worth flagging: this file merges into `develop` *cleanly* — no conflict — so it would land silently.

### 2. Generated `yarn set:dev` output was committed

Three separate production-reaching defects share one root cause. `yarn set:dev` runs `yarn set-all` (which rewrites the pbxproj) and then `scripts/dev-overrides.js`, and the resulting working-tree changes are in the commit.

**a. `android/app/src/main/res/xml/network_security_config.xml:3-4`** — `cleartextTrafficPermitted` flipped `false` → `true` on both `base-config` and `domain-config`. `scripts/dev-overrides.js:12-15` does exactly this `replaceAll`. The file is referenced from `src/main`, so this ships: it disables Android's HTTPS enforcement app-wide, in a cryptocurrency wallet, in release builds.

**b. `ios/BitPayApp.xcodeproj/project.pbxproj`, both app configs** — `ALLOWED_URL_PREFIXES = "ALLOWED-URL-PREFIXES-REPLACE-ME"` has been overwritten with a generated literal list in **both Debug and Release**. `develop` has 2 occurrences of that sentinel; this branch has 0. `scripts/allowed-url-prefixes.js` works by replacing the sentinel, so `yarn set:prod` is now a **silent no-op** and the allowlist is frozen at whatever your local `set:dev` produced. That list is missing the Sentry ingest domain (the script derives it from `SENTRY_DSN` at build time), so `AllowedUrlPrefixProtocol` will block Sentry in production.

**c. Same file, the app target's "Bundle React Native code and images" phase** — `export SENTRY_DISABLE_AUTO_UPLOAD=true` has been prepended, unconditionally, for every configuration. Every release archive stops uploading JS bundles and source maps, so production JS crashes arrive unsymbolicated. The workflow already sets this via `$GITHUB_ENV`; it doesn't need to be in the project file.

(b) and (c) independently blind production crash reporting, which is a bad thing to ship quietly.

**Fix:** `git checkout origin/develop -- ios/BitPayApp.xcodeproj/project.pbxproj android/app/src/main/res/xml/network_security_config.xml`, then re-apply only the deliberate Xcode changes (the new `BitPayAppUITests` target and its scheme entry). Please double-check `git status` before committing after any `yarn set:dev`.

### 3. `android/build.gradle:56-97` — `gradle.afterProject` rewrites packaging for the release APK

The `minSdkVersion` guard is fine (`:app` is 30, so it never fires). The `packagingOptions` block has no such guard and applies to every subproject including `:app`, so it reaches the release APK:

- `pickFirst` on `libhermes.so`, `libjsi.so`, `libreactnative.so`, `libc++_shared.so`, `libfbjni.so` — silences duplicate-native-library errors by picking an arbitrary copy. That error is the guard against an ABI mismatch between a vendored module's copy and `react-android`'s; silencing it globally converts a build failure into a runtime crash that only shows up on user devices.
- `excludes += ['META-INF/LICENSE', 'META-INF/NOTICE', 'META-INF/LICENSE.md', ...]` — strips third-party licence and notice files from the shipping APK. Apache-2.0 requires NOTICE redistribution, so this is a compliance problem, not just tidiness.
- `excludes += 'META-INF/*.kotlin_module'` — can break Kotlin top-level function resolution and reflection at runtime.
- `compileSdkVersion 36` forced onto any subproject declaring < 34, compiling third-party modules against an SDK they never declared.

If this is only needed to build `assembleDebugAndroidTest`, please scope it — `if (proj.name == 'app')` plus a variant filter, or move it into the androidTest configuration. A global `afterProject` hook that silently rewrites every module's packaging is a lot of blast radius for a test PR, and it's also one of the 14 merge conflicts (below).

### 4. A real BIP39 mnemonic is committed in plaintext

`hobby short divert lady spare quit act settle body town license alone` — a valid mainnet BIP39 phrase, in 5 Android test files and 1 iOS test file, in a public repo.

I checked: BIP44, BIP49 and BIP84 BTC addresses and the first three `m/44'/60'` ETH addresses are all empty right now, so nothing is at immediate risk — but:

- The wallet is permanently burned. If anyone funds it later (and `Test596BuyBTC` step 9 literally says *"tap My Key with funds available"*, so the suite expects it to be funded), those funds are spendable by anyone who reads this PR.
- It leaves a second way: `Test592ImportWalletRecoveryPhrase.kt` takes `allureScreenshot("Recovery phrase entered")` with the phrase on screen, and that screenshot is uploaded to Testlio and as a CI artifact.
- And a third: `android-espresso.yml` uploads the full `adb logcat` as the `device-logcat` artifact. The workflow's own comment at line 95 says RN dev logging serialises "every dispatched action" and that "a recovery-phrase import fires hundreds of actions". `DISABLE_DEVELOPMENT_LOGGING=true` is the only thing between the phrase and that artifact, on a debug build.
- It will trip secret scanners, and it can't be removed by a later commit.

Please treat it as compromised, generate a fresh fixture, and inject it via a GitHub secret rather than committing it. Same for `bc1q0effzahtsn685tyjppgukpvfhv37hrlm4g67ws`, the hardcoded send destination — that's a live mainnet address with 5 transactions and 181,650 sat received, now publicly linked to BitPay's test infrastructure.

### 5. The branch doesn't merge into `develop`

14 conflicts: `android/app/build.gradle`, `android/build.gradle`, `ios/BitPayApp/Info.plist`, `ios/Podfile.lock`, `package.json`, `src/lib/Braze/index.ts`, `src/navigation/wallet/screens/WalletDetails.tsx`, `AssetBalanceHistoryScreen.spec.tsx`, `portfolioChartVisibility.spec.tsx`, `app.effects.spec.ts`, `bitpay-id.effects.spec.ts`, `bitpay-id.effects.ts`, `wallet.spec.ts`, `wallet.reducer.spec.ts`.

The cause is `c75df96` / `7d5d0e9`, which merged `master` into a branch targeting `develop`. Relative to the merge base this branch **reverts** test coverage that is current on `develop`:

- `src/store/wallet/utils/wallet.spec.ts` — the whole `buildWalletObj` block (4 tests covering `tssKeyId`/`tssMetadata` propagation and the `readonly` keyId fallback)
- `src/store/bitpay-id/bitpay-id.effects.spec.ts` — 5 Cloudflare-challenge tests plus the `startLogin` describe, and the Braze merge-failure test

Those all conflict, so they can't land silently — but whoever resolves needs to take `develop`'s side on every one of them. Please merge `develop` (not `master`) into the branch and resolve before this is reviewable.

---

## Non-blocking, but please address

### Tests that can't fail

- **`ios/.../Test596_BasicBuyBTC_.swift:67-74`** — steps 10, 11 and 12 have **empty bodies**. The test's own description is "Best Offer → Continue With Opens Browser" and none of that is exercised or asserted. As written it passes whether or not offers load.
- **`ios/.../Test593_BTCConfirmPaymentScreen.swift:56`** — step "(7) Wait until the Send To screen is displayed" calls `isSendToTitleDisplayed()` and discards the result. No assertion.
- **`android/.../Test596BuyBTC.kt:90`** — `verifySelectCryptoTitleDisplayed()` result discarded. (Every other Android `verify*` call is correctly wrapped in `assertTrue` — this is the one exception.)
- **`android/.../Test596BuyBTC.kt:95-105`** — an entire `allureStep` block is commented out, so the buy test never selects a deposit wallet. Same at `Test594SwapBTC.kt:128`.
- **`android/.../pages/OnboardingPage.kt:194-224`** — `verifyIUnderstandCheckbox1Displayed()` returns `true` when post-onboarding state is visible, and `clickIUnderstandCheckbox1/2/3` and `clickAgreeAndContinue` all early-return before their assertions in the same state. If a regression causes the app to skip the mandatory terms-consent screen entirely, this suite goes green. The recovery behaviour is reasonable — it just needs to live in a separately named helper that can't satisfy an assertion.
- **`ios/.../Test595_BasicSellBTC_.swift`** — one assertion across six steps; step "(6) Enter $30 as the sell amount" enters `0.007`; the described outcome ("Continue With Opens Browser") is never asserted.

### `ios/BitPayAppUITests/flows/AppFlows.swift:78-82` — blind coordinate taps before money flows

When `goToHomeScreen` can't find the portfolio button, a Home tab, or a back control, it falls through to tapping absolute screen coordinate `(0.12, 0.11)` and dragging from the left edge — up to 15 times, with no knowledge of what screen it's on. This runs in `setUp()` for the send, swap, sell and buy tests, i.e. immediately before the flows that move money. (The element-relative `element.coordinate(...)` calls elsewhere in the page objects are fine — this is the only absolute one.)

It also never fails: after `attempts` it returns silently, so the test proceeds from an unknown screen and the eventual failure points at the wrong step. An `XCTFail` at the end would make these much easier to debug.

### iOS tests aren't isolated (Android's are)

`scripts/run-tests.sh` does `adb shell pm clear` before each Android class, and each Android money-flow test re-imports the fixture itself. The iOS side does neither: only `Test592` imports the phrase, and `Test593`–`596` call `completeOnboardingIfRequired`, which on clean state creates a **brand-new empty random wallet** rather than importing the fixture. So the iOS money-flow tests either inherit whatever `Test592` left behind or run against an empty wallet, and none of them can be run individually. Given the PR is specifically about the iOS suite, this is worth fixing before merge.

### Workflows

- **Duplicate workflow names.** `ios-xcuitest.yml` and `ios-xcuitest-sandbox.yml` are both `name: iOS XCUITest`; both Android files are both `name: Android Espresso E2E`. `android-espresso.yml` then triggers on `workflow_run: workflows: ["iOS XCUITest"]`, which matches either — so a manual sandbox run kicks off the Android job using the **production** Testlio token and project.
- **No success guard.** That same `workflow_run` has `types: [completed]` with no `github.event.workflow_run.conclusion == 'success'` check, so the Android suite runs after a failed or cancelled iOS run.
- **`workflow_run` checks out the default branch.** `android-espresso.yml:21` is a bare `actions/checkout@v4`; under `workflow_run` that's `master`, not the ref the triggering run tested. Relatedly, `schedule:` only fires on the default branch, so the nightly won't run at all until this reaches `master`.
- **Colliding concurrency groups.** Both iOS workflows use `e2e-ios-xcui-${{ github.ref }}` and both Android ones use `e2e-android-espresso-${{ github.ref }}`, with `cancel-in-progress: true` — a manual run cancels the nightly.
- **"Sandbox" doesn't sandbox anything.** The only difference between the two variants is which Testlio workspace receives the report. Both build identically (`cp .env.template .env.development` + `yarn set:dev`), and `src/constants/config.ts:41` hardcodes `APP_NETWORK = Network.mainnet` with no env override, so both run on **mainnet against `https://bitpay.com` and `https://bws.bitpay.com`** and hit live Changelly/Moonpay endpoints nightly. The naming implies an isolation that doesn't exist — please either rename it or wire the input through to the app config. (The send test does stop at the Confirm Payment summary and never slides to send, which is the right call.)
- **Broken logcat pid capture.** `android-espresso.yml:167-172`: `android-emulator-runner` splits `script:` per line and runs each as a separate `sh -c`, so `adb logcat ... &` and `echo $! > /tmp/logcat.pid` are in different shells, `$!` is unset, and `kill $(cat /tmp/logcat.pid) || true` is a silent no-op. (Metro's pid capture is fine — it's inside one `run: |` block. And test failures *do* propagate correctly; I checked the action's source.)
- **Secrets inline in `run:` bodies.** `echo "${{ secrets.SENTRY_PROPERTIES }}" > sentry.properties` and `export RUN_API_TOKEN="${{ secrets.TESTLIO_RUN_TOKEN }}"` splice secret values into the script text. Please pass them via `env:` instead. Also: Android writes real Sentry properties, iOS just `touch`es empty files — one of the two is wrong.
- **No `permissions:` block** on any of the four new workflows.
- **Empty results upload cleanly.** `run-tests.sh` `adb pull ... || true`, then `cp ... || true`, then a "Check Allure results" step that emits `::warning::` and exits 0 — so total loss of results is a green step that uploads an empty archive to Testlio.
- ~540 lines of near-duplicate YAML across each sandbox/scheduled pair; one workflow with an input would be much easier to keep in sync.

### Timeouts and retries

- **`RetryRule` never retries.** `BaseTest.kt:37` documents "Defaults to 3"; `BaseTest.kt:50` is `retryCount: Int = 1`; `RetryRule.kt:8` does `repeat(retryCount)`, so `1` means one attempt. The rule, the ordering comment, and the screenshot-on-final-failure design are all inert as configured. Either fix the default or drop the machinery.
- **Very long single-element waits.** `MyKeyPage.kt:25` is 900,000 ms (15 min); `OnboardingPage.kt:93,104,162` are 600,000 ms; `:199` is 420,000 ms. `WaitUtils.waitForView` catches `Throwable`, so if the app has crashed it retries for the full window instead of failing fast — a couple of those and the 75-minute emulator step is gone before anything reports.

### Xcode project

- **`RCTCrashTest` removed from the app target.** `develop` has 6 references in the pbxproj; this branch has 0, while `ios/RCTCrashTest.mm` is still in the tree. `src/lib/crash-test/index.ts` will fail with "native module unavailable". Looks unintentional — probably the same pbxproj regeneration as §2.
- **`ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = "$(inherited)"` deleted** from both app configs (2 occurrences on `develop`, 0 here). Also looks unintentional.
- **`PRODUCT_BUNDLE_IDENTIFIER = com.test.appium.BitPayAppUITests`** — template leftover; should be under a BitPay identifier.
- **`IPHONEOS_DEPLOYMENT_TARGET = 26.0`** on the test target is what pins CI to Xcode 26.2 and iPhone 17 only.
- **`TestPlan.xctestplan` is dead config** — the scheme has no `TestPlanReference`, neither workflow passes `-testPlan`, and the file is in *Copy Bundle Resources* (`project.pbxproj:423`), so it ships inside the `.xctest` bundle as a resource rather than configuring anything.
- **`parallelizable = "YES"`** on the new `TestableReference` while the suite is order-dependent and CI passes `-parallel-testing-enabled NO`. Anyone running the scheme from Xcode gets a broken suite.
- **`systemAttachmentLifetime`/`userAttachmentLifetime = "keepAlways"`** in the shared scheme makes every developer's local test run keep all attachments forever.

### Smaller things

- **`MainActivity.kt:51`** — gating `FLAG_SECURE` on `!BuildConfig.DEBUG` is scoped to debug and doesn't affect release, so I'm fine with it in principle, but it does mean debug builds can screenshot the recovery-phrase screen. Worth a deliberate nod from someone rather than arriving inside a test PR.
- **`README.md:26-31`** — the `network_security_config` example got mangled into invalid XML (`<domain-config></domain>`, `<trust-anchors>` merged into the closing tag, the `<certificates>` line lost). There's also a stray `# Tests` H1 at the top of the file.
- **Android tests carry no `testlioManualTestID`** — all 6 iOS tests set it, 0 of 6 Android tests do, so the Android results won't link to their manual cases in Testlio.
- **Android and iOS aren't testing the same artifact** — Android builds `assembleDebug` and serves JS from a live Metro bundler over `adb reverse` (`__DEV__` true, unminified); iOS builds `-configuration Release` against a pre-generated `--dev false` bundle.
- **`VerifyYourPhrasePage.swift` and `NewRecoveryPhrasePage.swift` are dead** — not referenced by any test or flow. If they're kept: `extractNumbers` matches `\d+` against instructions that repeat each index ("2nd (2)"), yielding six indexes for three fields and then indexing past the end of `textFields`; there's also a force-unwrapped `Range(...)!` and a hardcoded `.prefix(12)`.
- **`SwapPage.swift:13` / `SellPage.swift:13`** wait on the global "Swap"/"Sell" static texts, which `LinkingButtons.tsx` already renders on Home — so a failed navigation satisfies the wait immediately and the test fails later in a misleading place.
- **`scripts/run-tests.sh`** hardcodes `/home/runner/work/_temp/allure-results`, so it can't run locally despite living in `scripts/`. It's also mode `100644` (fine as long as it's always invoked via `bash`).
- Missing trailing newlines on `android/build.gradle`, `scripts/run-tests.sh`, all four workflows and `androidTest/AndroidManifest.xml`.
- `AppFlows.swift:5` carries a personal name in the file header.

### Relayed, not verified by me

`androidTestImplementation "io.qameta.allure:allure-kotlin-*:2.4.0"` (`android/app/build.gradle:181`) — flagged as having a path where an uncaught app `RuntimeException` is recorded as a *passing* result ([allure-kotlin#81](https://github.com/allure-framework/allure-kotlin/issues/81)). I haven't reproduced it on an emulator, so treat it as a lead. If it holds, a crash of the app under test could be reported to Testlio as green, which would be worth pinning around.

---

android/app/src/main/java/com/bitpay/wallet/MainActivity.kt
TestlioHaritha: This change is intentionally limited to debug builds to support Espresso/Appium automation and Layout Inspector screenshots. Release builds will retain FLAG_SECURE.

Could you please direct me to the appropriate Android/app security owner for explicit approval? I’ll wait for their confirmation before proceeding.

android/app/src/androidTest/java/com/bitpay/wallet/utils/RetryRule.kt
TestlioHaritha: Updated

android/app/src/androidTest/java/com/bitpay/wallet/pages/OnboardingPage.kt
TestlioHaritha: updated

ios/BitPayAppUITests/tests/Test596_BasicBuyBTC_.swift
TestlioHaritha: updated

ios/BitPayAppUITests/tests/Test593_BTCConfirmPaymentScreen.swift
TestlioHaritha: updated

android/app/src/main/java/com/bitpay/wallet/MainActivity.kt
JohnathanWhite: disregard this one sorry

.github/workflows/android-espresso.yml
TestlioHaritha: We have observed instances where one platform test can fail for various reasons. To avoid blocking Android execution due to an iOS test failure, we intentionally did not add an explicit condition requiring iOS success.

SaifTestlio: Updated

.github/workflows/ios-xcuitest-sandbox.yml
SaifTestlio: I have updated the Run tokens to env:
Though it is for use of results to upload to a different sandbox workspace. It is not needed to app env changes.
CC: @TestlioHaritha

PR #2160, opened 2026-05-06
Theme: Fix - Replace `null` with `unspecified` to standardize system default handling

cmgustavo:
RN-2674

### Changes

* Replaces all usages of null as the "system default" color scheme value with 'unspecified', aligning with the updated **ColorSchemeName** type in React Native ('light' | 'dark' | 'unspecified')
* Updates initial state, resetAllSettings, wallet import migration, the Theme screen UI, analytics/log labels, and the appearance change listener
* Adds backward-compatible null handling at all read sites (Root.tsx, index.js, Theme.tsx) so existing users with null persisted in storage continue to get correct system-default behavior after upgrading

WhatsApp Image 2026-05-02 at 11 15 24 PM

### DIFF
--- a/index.js
+++ b/index.js
@@ -205,7 +205,7 @@ const AppWrapper = () => {
 
     const subscription = Appearance.addChangeListener(
       ({colorScheme: newScheme}) => {
-        if (colorScheme === null) {
+        if (!colorScheme || colorScheme === 'unspecified') {
           setIsDark(newScheme === 'dark');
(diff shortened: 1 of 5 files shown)


gabrielbazan7 (approved):
ACK !

JohnathanWhite:
Synced to Jira: https://bitpayprod.atlassian.net/browse/RN-2808

PR #2164, opened 2026-05-11
TSS: Fix - Cancel ceremony when unmounting mid-join (new-join path)

cmgustavo:
Added `cancelOnCreateRef` to signal deferred cancellation: if the component unmounts before the keyId is known, the flag is set, and the first thing `onKeyCreated` does when it eventually fires is dispatch `cancelTSSCeremony` — closing the orphan immediately


Gamboster (approved):
LGTM

JohnathanWhite (approved):

PR #2166, opened 2026-05-12
Portfolio: Fix - Hide balance chart and asset list loaders on fresh install

msalcala11:
Balance charts and asset lists now only appear when at least one wallet has a non-zero balance. Also ensures portfolio populate tx history requests pass the currently installed BWC version.


cmgustavo (approved):
ACK!

PR #2172, opened 2026-05-25
Buy: Fix - Unhide EVM account when a new wallet is created via Buy Crypto

cmgustavo:
When buying a token (e.g. USDC on Ethereum) whose parent EVM account was previously hidden, the newly created wallet would appear visible while the rest of the account (ETH, Polygon, Arbitrum, etc.) remained hidden, causing an inconsistent state on the home screen.

Fix: after addWallet() succeeds in the Buy Crypto flow, check if the created wallet's receive address belongs to a hidden EVM account and if so, unhide the entire account automatically.


gabrielbazan7 (approved):
LGTM

PR #2175, opened 2026-05-27
CreateAccount: Fix - Use crypto.randomBytes for salt generation

cmgustavo:
Replace PBKDF2-based salt generation with crypto.randomBytes, consistent with the approach already used in `pin.ts`.

The previous implementation could pass 0 iterations to pbkdf2Sync and blocked the JS thread unnecessarily. This PR doesn't affect current users.

### DIFF
--- a/src/utils/password.ts
+++ b/src/utils/password.ts
@@ -1,4 +1,5 @@
 import Pbkdf2 from 'pbkdf2';
+import crypto from 'crypto';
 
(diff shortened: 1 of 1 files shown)


gabrielbazan7 (approved):
ACK

JohnathanWhite:
Synced to Jira: https://bitpayprod.atlassian.net/browse/RN-2810

PR #2176, opened 2026-05-27
Portfolio: Enhancement - Consolidate portfolio worklet runtimes to reduce RAM usage

cmgustavo:
Replace three separate WorkletRuntime instances (main, rates, analysis) with a single shared runtime.

Also fixes resetPortfolioRuntimeClient which was not resetting the main runtime reference.


cmgustavo:
### Memory Pressure: 
* Before fix: 1,104,402 KB (~1.08 GB) 
* After fix: 1,074,765 KB (~1.05 GB) 
* Difference: ~29 MB

### Private Dirty: 
* Before fix: 904,276 KB (~883 MB) 
* After fix: 865,212 KB (~845 MB) 
* Differente: ~38 MB

The fix saves ~29–38 MB, that is not sufficient to prevent the WatchdogTermination. iOS crashes the app at roughly 1.2–1.5 GB on iPhone 11 Pro. So, PR is not solving current issue.

PR #2196, opened 2026-06-09
Fix: Check ABI filters before build for old Android devices

cmgustavo:
## Problem
The SIGABRT crash caused by missing `armeabi-v7a` ABI filters has happened twice (14.39.15 and 14.43.2). Both times the root cause was the same: `armeabi-v7a` missing from `android/app/build.gradle`, affecting mid/low-end Android devices running in 32-bit mode.

## Solution
Add a `postinstall` script that validates `armeabi-v7a` is present in all 3 required locations in `android/app/build.gradle`:

- `reactNativeArchitectures()` default return value
- `ndk { abiFilters }` block
- `versionCodes` map

### DIFF
--- a/scripts/check-abi-filters.js
+++ b/scripts/check-abi-filters.js
@@ -1,0 +1,52 @@
+// scripts/check-abi-filters.js
+const fs = require('fs');
+const path = require('path');
+
+const gradlePath = path.join(process.cwd(), 'android', 'app', 'build.gradle');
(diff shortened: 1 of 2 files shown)


JohnathanWhite:
Synced to Jira: https://bitpayprod.atlassian.net/browse/RN-2827

JohnathanWhite (approved):

PR #2214, opened 2026-06-19
Wallet: Fix - Crash adding custom token to read-only key

cmgustavo:
Adding a custom token (or a Solana wallet via AddingOptions) crashed with "Cannot read property 'addKeyByAlgorithm' of undefined" when the key had no live methods instance (read-only/watch-only keys, where buildKeyObj sets `methods: undefined`)


JohnathanWhite:
Synced to Jira: https://bitpayprod.atlassian.net/browse/RN-2842

JohnathanWhite (approved):

PR #2222, opened 2026-07-06
Sentry: Feat - Add DEV-only crash test buttons

cmgustavo:
Adds a `__DEV__` "Crash Test" section to the About settings screen with buttons to trigger native and JS crashes, so crash reporting to Sentry can be validated end-to-end on-device across all platforms


JohnathanWhite:
https://bitpayprod.atlassian.net/browse/RN-2859

JohnathanWhite (approved):

PR #2225, opened 2026-07-13
Config: Fix - Resolve implicit-any on network-indexed maps

cmgustavo:
Fixes TS7053 (`Element implicitly has an 'any' type because expression of type 'any' can't be used to index type ...`) reported when indexing `BASE_BITPAY_URLS` and `METHOD_ENVS` with a network value that resolves to `any` at the call site


JohnathanWhite:
https://bitpayprod.atlassian.net/browse/RN-2862

JohnathanWhite (approved):

PR #2231, opened 2026-07-17
Portfolio: Fix - Prevent worklet cached Hermes value race

msalcala11:
The react-native-worklets package let multiple Worklet threads replace the same cached Hermes value at once, causing one thread to destroy memory another thread was still using, which led to a crash. This patch locks cache ownership and makes other runtimes create their own temporary values instead.


JohnathanWhite:
https://bitpayprod.atlassian.net/browse/RN-2868

JohnathanWhite (approved):

PR #2234, opened 2026-07-22
Buy/Sell: Fix - params update for ramp network signed payment url generation

Gamboster:
Ticket reference: [RN-2486](https://bitpayprod.atlassian.net/browse/RN-2486)

Ramp deprecated the legacy per-flow search params in favor of a unified format.
Ref: https://docs.rampnetwork.com/search-params-migration
Older app versions still send the legacy fields, so we translate them into the new unified params in BWS (With this PR: https://github.com/bitpay/bitcore/pull/4204) to keep those requests working going forward. If a newer client already sends the new params, those take precedence.

The `useSendCryptoCallbackVersion: 1` param was removed, as it was causing errors in Offramp integration.


JohnathanWhite:
https://bitpayprod.atlassian.net/browse/RN-2869

JohnathanWhite (approved):

PR #2241, opened 2026-07-24
App Performance: Enhancement - remove styled-components and optimize rendering performance

gabrielbazan7:
**Performance diagnostics (`PERF_DEBUG`)**

All performance instrumentation is gated behind a single constant in `src/utils/performanceDebug.ts`, off by default. Set it to `true` and reload to enable it; logs go to the Metro console.

| Prefix | What it measures |
| --- | --- |
| `[PERF-NAV]` | screen transitions — `transitionDeltaMs` is the animation alone, `stateDeltaMs` also covers the work before it starts (forward navigations only) |
| `[PERF-SCREEN]` | a screen's mount timeline: commit → layout → first frame → second frame |
| `[PERF-REACT]` | React render/commit per subtree — compare `actualMs` against `baseMs` to spot missing memoization |
| `[PERF-REDUX]` | dispatch and reducer durations |
| `[PERF-PERSIST]` | redux-persist writes to MMKV |

`[PERF-NAV]`, `[PERF-REDUX]` and `[PERF-PERSIST]` work as soon as the flag is on. `[PERF-SCREEN]` needs `useScreenRenderPerformance(name)` wired to the screen's `onLayout`, and `[PERF-REACT]` needs the subtree wrapped in `<PerformanceProfiler id="..." onRender={logReactProfiler}>` — both are no-ops while the flag is off, so they're safe to leave in place.

Keep the flag `false` on merge: `performanceLog` is a plain `console.log`, and `performanceDebug.spec.ts` asserts the default so an accidental commit fails CI.


src/components/styled/Containers.tsx
```diff
@@ -31,63 +38,501 @@ export const isNarrowHeight = HEIGHT < 700;
...
+
+export const HeaderTitleContainer = React.forwardRef<
+  View,
+  React.ComponentProps<typeof View>
+>(({style, ...rest}, ref) => (
+  <View ref={ref} style={[styles.headerTitleContainer, style]} {...rest} />
+));
+HeaderTitleContainer.displayName = 'HeaderTitleContainer';
+
+export const ScreenContainer = React.forwardRef<
```
JohnathanWhite, line 499: `ScreenContainer` was `styled.SafeAreaView` on develop (`Containers.tsx:50`); here it becomes a plain `View` with `styles.screenContainer` = `{flex: 1}` (:124). `OptionContainer` has the identical change at :974 (was `styled.SafeAreaView` at :349 on develop).

Nothing upstream compensates for this. The app's only root-level safe area is `Root.tsx:751`, and it's `edges={['left', 'right']}` — top and bottom insets are deliberately left to the screens. So every screen relying on these two containers will now draw content under the status bar / notch and under the home indicator.

That's 21 files using `ScreenContainer` and 20 using `OptionContainer`.

Fix is to keep them safe-area-aware — either `SafeAreaView` from `react-native-safe-area-context`, or `useSafeAreaInsets()` with the padding applied — rather than a bare `View`. Worth a quick device check on a notched iPhone and a gesture-nav Android afterward.

src/store/wallet/utils/accountListCache.ts
```diff
@@ -0,0 +1,494 @@
...
+
+  return `${keys.length}:${wallets.length}:${hash}`;
+};
+
+const storeSnapshotInMemory = <T>(
+  cacheKey: string,
+  signature: string,
+  value: T,
+): void => {
+  snapshots.delete(cacheKey);
```
JohnathanWhite, line 418: `storeSnapshotInMemory` does `snapshots.delete(cacheKey)` immediately followed by `.set(...)` — that's the standard LRU recency-bump idiom, but there's no size cap, so nothing is ever actually evicted. `snapshots` (:57) grows for the life of the process; the only thing that releases it is `clearAccountListSnapshots()` (:477), called from `walletDerivedCacheLifecycle.ts:22`.

Worth comparing against the sibling cache added in this same PR: `balanceHistoryChartSeriesCache.ts:877` does exactly this pattern *with* `MAX_CACHED_BALANCE_CHART_SERIES = 50` and drops the oldest key. So the intent clearly exists — it just didn't make it into this module.

In practice the key is per key-id + account address + context, so it's bounded by how many accounts the user actually has rather than unbounded in the scary sense. But the values are fully-built row lists, and the persisted side only caps *per entry* at 256 KiB (:225), not in total. Adding the same cap here would make the two caches consistent and put a ceiling on it.

src/components/charts/useBalanceChartDisplayModel.ts
```diff
@@ -383,23 +468,33 @@ export function useBalanceChartDisplayModel({
...
+        selectedTimeframeIsVisible ||
+        (preserveVisibleSeriesWhileNotReady && visibleOwnerMatchesScope);
 
       activeRequestIdRef.current += 1;
       pendingSelectedTimestampRef.current = undefined;
       gestureStartedRef.current = false;
       lastHapticPointTsRef.current = undefined;
       setSelectedPoint(undefined);
-      if (!canPreserveVisibleSeries) {
+      if (!canPreserveVisibleSeries && !preserveVisibleSeriesWhileNotReady) {
```
JohnathanWhite, line 494: In this `!isBalanceChartDataReadyToQuery` branch: when a cached series matches the current scope and timeframe, `canPreserveVisibleSeries` is true, the cached series is re-shown, and `setLoading(!selectedTimeframeIsVisible)` evaluates to `false`.

The effect is that a previous session's chart is presented as current — no spinner, no staleness affordance — for as long as the data never becomes ready to query, up to the 48h TTL (`balanceHistoryChartSeriesCache.ts:44`). Stale-while-revalidate is the right pattern here, but this is the branch where the revalidate leg is precisely the thing that isn't running, and the user has no way to tell.

Suggest either keeping a subtle loading/stale indicator while in this branch, or bounding how old a preserved series may be before it's dropped rather than leaning on the full 48h.

To be clear about what is *not* wrong here, since it looks suspicious at first glance: the lookup key at `balanceHistoryChartSeriesCache.ts:89` does omit the rate and chart-data revisions, but entries carry `queryRevisionKey` / `seriesSignature` and the normal path does replace them correctly. This is specifically about the not-ready branch, not a broken cache key.

package.json
```diff
@@ -201,7 +201,6 @@
     "rn-swipe-button": "3.0.1",
     "safe-json-utils": "1.1.1",
     "stream-browserify": "1.0.0",
-    "styled-components": "5.3.3",
     "text-encoding": "0.7.0",
```
JohnathanWhite, line 204: `@shopify/flash-list` 2.0.3 -> 2.3.2, and across the branch FlashList usage roughly doubles (48 -> 95 references).

Both are reasonable changes on their own merits, but they're a list-rendering-engine change riding inside a styling PR, and they're the part most likely to cause a regression that's hard to attribute after the fact. Lockfile is updated and I don't see any use of the v1 `estimatedItemSize` API, so nothing looks broken.

No action needed if it's deliberate — mainly flagging so it gets its own QA pass, and so that if something list-related regresses after this lands, this is the first thing to look at rather than the style conversion.

JohnathanWhite (changes requested):
Went through this focusing on the parts that aren't mechanical — the two new cache modules, the shared primitives, and the five most-rewritten screens.

The conversion itself is disciplined, and worth saying so: all 349 `StyleSheet.create` calls are module-level (none recreated per render), there are zero residual `styled-components` imports, and the dependency, both `@types` packages, the patch file and `styled.d.ts` are all cleanly removed. Theme handling got *broader*, not narrower (`useTheme()` 224 -> 859 references), and index-based `keyExtractor`s went down (13 -> 9). A dependency/memoization pass over GlobalSelect, AccountDetails, KeyOverview, WalletDetails and HomeRoot came back clean.

Four things worth addressing, one of which I think blocks:

1. **`ScreenContainer` and `OptionContainer` lost their safe-area behavior** — 41 files affected. Details inline.
2. `accountListCache` has no eviction, where its sibling cache does.
3. The balance chart can present a stale series as current when the query can't run.
4. The FlashList bump is unrelated scope riding along.

Details inline on each.

src/components/styled/Containers.tsx
gabrielbazan7: Done!

src/components/charts/useBalanceChartDisplayModel.ts
gabrielbazan7: Done!

src/store/wallet/utils/accountListCache.ts
gabrielbazan7: Done

src/store/wallet/effects/status/status.ts
```diff
@@ -460,22 +470,60 @@ export const startUpdateAllWalletStatusForKeys =
...
       try {
         logManager.info('starting [startUpdateAllWalletStatusForKeys]');
         const keyUpdatesPromises = keys.map(key =>
-          dispatch(updateKeyStatus({key, accountAddress, force})),
+          dispatch(
+            updateKeyStatus({
+              key,
+              accountAddress,
+              force,
+              dataOnly: true,
```
msalcala11, line 483: With `dataOnly: true`, successful requests no longer update Redux themselves. If any key's request fails, `Promise.all(keyUpdatesPromises)` at line 487 throws before the batch dispatch, which prevents successful updates for other keys from being applied.

Replacing that specific call with `Promise.allSettled(keyUpdatesPromises)`, extracting the fulfilled values, and passing them to the existing `successUpdateWalletBalancesAndStatus` dispatch before reporting failures preserves successful updates. A regression test with one successful key and one failed key should verify that the successful key's balances and proposals still reach Redux.

msalcala11 (changes requested):

src/store/wallet/effects/status/status.ts
gabrielbazan7: Thanks!

PR #2126, opened 2026-04-08
Transaction List: Fix - Month name localization for regional locales (e.g. zh-cn)

cmgustavo:
Fix issue reported on Android:
```
Requiring unknown module "./locale/zh"
```

This was not related to our i18n translations, but to Moment.js locale loading behavior.

---

### Month Name in English (should be zh)

IMG_1208

### Fixed

IMG_1207

---
[RN-2652](https://bitpayprod.atlassian.net/browse/RN-2652)

PR #2262, opened 2026-08-04
Transaction List: Fix - Fallback to 'en' for locales moment doesn't bundle

cmgustavo:
PR #2126 only special-cased zh -> zh-cn; any other unsupported locale
(such as 'co' Corsican) still reached moment().locale().

Fix: Validate the candidate against moment.locales() and fall back to 'en' when it's not bundled.


Gamboster (approved):
LGTM

PR #2264, opened 2026-08-05
bump pbkdf2 iterations to owasp rec

kajoseph:
Bumps the p2kdf2 iterations to the OWASP recommendation of 600k

### DIFF
--- a/src/navigation/wallet/screens/wallet-settings/ExportWallet.tsx
+++ b/src/navigation/wallet/screens/wallet-settings/ExportWallet.tsx
@@ -169,7 +169,7 @@ const ExportWallet = () => {
     backup = JSON.stringify(backup);
 
     return JSON.stringify(
-      Encryption.encryptWithPassword(backup, password, {iter: 1000}),
+      Encryption.encryptWithPassword(backup, password, {iter: 600_000}),
     );
   };
 
--- a/src/store/wallet/effects/status/status.ts
+++ b/src/store/wallet/effects/status/status.ts
@@ -34,7 +34,7 @@ import {createWalletAddress} from '../address/address';
 import {detectAndCreateTokensForEachEvmWallet} from '../create/create';
 import uniqBy from 'lodash.uniqby';
 import {logManager} from '../../../../managers/LogManager';
-import {type BulkClient} from '@bitpay-labs/bitcore-wallet-client/ts_build/src/lib/bulkclient';
+import {type BulkClient} from '@bitpay-labs/bitcore-wallet-client/src/lib/bulkclient';
 export const startUpdateWalletStatus =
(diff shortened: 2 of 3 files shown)


src/navigation/wallet/screens/wallet-settings/ExportWallet.tsx
```diff
@@ -169,7 +169,7 @@ const ExportWallet = () => {
     backup = JSON.stringify(backup);
 
     return JSON.stringify(
-      Encryption.encryptWithPassword(backup, password, {iter: 1000}),
+      Encryption.encryptWithPassword(backup, password, {iter: 600_000}),
```
JohnathanWhite, line 172: **600× more KDF work, synchronously on the JS thread**

`encryptWithPassword` is called synchronously here and at `ExportTSSWallet.tsx:249`, so raising `iter` from 1,000 to 600,000 puts 600× the key-derivation work on the React Native JS thread with no yield. On lower-end Android hardware that's a plausible ANR, and at minimum a multi-second freeze during wallet export with no progress indication.

The OWASP alignment is worth keeping — the issue is only the thread it runs on. Moving it to native/background work, or chunking with a progress state, would keep the stronger work factor without the freeze.

Backwards compatibility checks out, for the record: `encryptWithPassword` records `iter` in the envelope and `decryptWithPassword` reads `json.iter` back, so existing exports created at 1,000 iterations still decrypt.

src/store/wallet/effects/status/status.ts
```diff
@@ -34,7 +34,7 @@ import {createWalletAddress} from '../address/address';
 import {detectAndCreateTokensForEachEvmWallet} from '../create/create';
 import uniqBy from 'lodash.uniqby';
 import {logManager} from '../../../../managers/LogManager';
-import {type BulkClient} from '@bitpay-labs/bitcore-wallet-client/ts_build/src/lib/bulkclient';
+import {type BulkClient} from '@bitpay-labs/bitcore-wallet-client/src/lib/bulkclient';
```
JohnathanWhite, line 37: **Unrelated import-path change in a KDF PR**

This swaps `@bitpay-labs/bitcore-wallet-client/ts_build/src/lib/bulkclient` for `.../src/lib/bulkclient`, pointing the type import at the package's TypeScript sources rather than its build output. It's type-only so it erases at compile time, and Types is green.

But it's unrelated to bumping pbkdf2 iterations, and it makes this the only such import in the codebase: 6 files under `src/` import from `ts_build/`, none from `src/`. Worth splitting out or reverting to keep the PR to one concern.

JohnathanWhite (commented):

Two findings. The OWASP change itself looks right — the concerns are about where the work runs, and one unrelated edit.

JohnathanWhite (approved):

PR #2267, opened 2026-08-05
Guidepoint: Feat - Add root/jailbreak detection (defense-in-depth)

cmgustavo:
Jira: [RN-2912](https://bitpayprod.atlassian.net/browse/RN-2912)

Adds client-side root (Android) / jailbreak (iOS) detection. Centralized in utils/deviceIntegrity.ts via jail-monkey 3.0.0
(New Architecture-compatible). 

Tiered, non-blocking response:

- Silent: tag Sentry (device.compromised) on every event + analytics on detection
- Soft block: passkey register/auth and crypto signing on compromised devices; passkey login falls back to basic auth, no full-app lockout

### Testing (no root needed)

Debug build → press j in Metro to open React Native DevTools → Console tab:

```
globalThis.__deviceIntegrity.simulate('jailbroken')  // or 'hook-detected'
globalThis.__deviceIntegrity.clear()                 // reset
``` 

With the override on, verify:

* Sign a tx / register passkey → blocked with the rooted/jailbroken message
* Passkey login → falls back to password (doesn't fail)
* Sentry → tagged device.compromised=true
* Rest of the app → still works (no hard block)

Note: the one-time Device Integrity Compromised analytics event only fires at app init — to test it, use a release build on an emulator that trips detection.

Example error:

```
[Warn] [deviceIntegrity] sensitive action "crypto-sign" attempted on compromised device (reason: jailbroken)
```

IMG_1766


src/utils/deviceIntegrity.ts
```diff
@@ -0,0 +1,157 @@
...
+// Dev-only override to simulate a compromised device (see global seam below).
+let devOverride: DeviceIntegrityResult | null = null;
+
+export const getDeviceIntegrity = (
+  forceRefresh = false,
+): DeviceIntegrityResult => {
+  if (__DEV__ && devOverride) {
+    return devOverride;
+  }
+  if (cached && !forceRefresh) {
```
JohnathanWhite, line 36: **Session-long cache means `hookDetected` is only ever evaluated once — and a failed check caches CLEAN**

Two problems, both reachable through `assertDeviceIntegrityForSensitiveAction`, which calls `getDeviceIntegrity()` with no `forceRefresh`:

1. The comment above the cache says state can't change without a reboot. That holds for root/jailbreak, but not for `hookDetected` — a Frida-style hook can attach *after* startup, and `hookDetected` is folded into `isCompromised`. Because the startup telemetry call populates the cache, every later signing check reuses the pre-hook result.
2. The `catch` assigns `cached = CLEAN` (line 55). A single transient native failure at startup therefore marks the device clean for the whole session. Fail-open is a defensible choice; caching the fail-open result makes it permanent.

Calling `getDeviceIntegrity(true)` from the sensitive-action path, and leaving the catch result uncached, would address both.

src/store/bitpay-id/bitpay-id.effects.ts
```diff
@@ -270,6 +271,14 @@ export const checkLoginWithPasskey =
         throw err;
       }
 
+      // Compromised device: fall back to basic auth instead of failing login.
+      if (isDeviceCompromisedError(err)) {
+        logManager.warn(
+          '[checkLoginWithPasskey] passkey auth blocked on compromised device, falling back to basic auth',
+        );
+        return Promise.resolve(false);
```
JohnathanWhite: **Detecting a compromised device downgrades auth rather than blocking it**

`isDeviceCompromisedError(err)` returns `Promise.resolve(false)`, and `false` is the same value the caller receives when passkeys are simply unavailable. So on a rooted or hooked device the net effect is: passkey auth — the stronger factor — is removed, and the user is steered to basic auth on that same compromised device.

That may be the intended availability trade-off, but it's worth stating explicitly in a PR titled *defense-in-depth*: as written, the detection weakens the auth path it fires on. If blocking is intended, propagate the error and surface the blocking message instead of collapsing it into the "no passkey" signal.

JohnathanWhite (commented):

Two findings, one of them worth a look before this lands.

src/utils/deviceIntegrity.ts
cmgustavo: Good catch, both fixed. The catch now returns CLEAN without caching it, and the sensitive-action path calls `getDeviceIntegrity(true)` so hookDetected is re-evaluated per signing attempt. The cache is now only for cheap reads (startup telemetry, useDeviceIntegrity). 
Tests added for both.

src/store/bitpay-id/bitpay-id.effects.ts
cmgustavo: Blocking `signInWithPasskey()` never prevented login (it fell through to basic auth), so all it did was remove the stronger factor on exactly the device where it matters. 
Removed the check from `signInWithPasskey()`; the passkey is held in Secure Enclave / StrongBox and remains the best available factor even on a compromised OS. `registerPasskey()` stays blocked — binding a new credential to a device we already distrust is a different call. The fallback branch in checkLoginWithPasskey is now dead code and deleted.

JohnathanWhite (approved):

PR #2278, opened 2026-08-10
Encryption: Fix - replace persisted CBC encryption with AES-GCM

gabrielbazan7:


src/store/transforms/encrypt.ts
```diff
@@ -176,12 +368,16 @@ const transformShopStore = (
...
     const updatedCard = {...card};
     fieldsToTransform.forEach(field => {
       const value = card[field];
-      if (value && typeof value === 'string' && checkCondition(value)) {
-        updatedCard[field] = transformer(value, secretKey);
+      if (hasProtectedValue(value) && checkCondition(value)) {
+        updatedCard[field] = transformer(
+          value,
+          secretKey,
+          `SHOP.giftCards.${Network.mainnet}.${cardIndex}.${field}`,
```
JohnathanWhite, line 379: **Gift-card AAD binds ciphertext to its array index, which is not stable**

The AAD for gift-card fields is `SHOP.giftCards.${Network.mainnet}.${cardIndex}.${field}`, so each ciphertext is bound to the card's **position** in the array. That position is not stable: `DELETED_UNSOLD_GIFT_CARDS` calls `.filter(card => card.status !== 'UNREDEEMED')` on the array (`src/store/shop/shop.reducer.ts`), and `SET_PURCHASED_GIFT_CARDS` replaces it wholesale with a server-ordered list. Because `encryptShopStore` skips values that are already encrypted (`!isEncryptedValue(value)`), the surviving cards keep ciphertext whose AAD encodes their *old* index, while `decryptShopStore` recomputes the AAD from the *new* index — so `decipher.final()` fails GCM tag verification. With `transforms.ts` now rethrowing, that single failure aborts rehydration of the whole persisted root. Consider binding the AAD to something stable and card-intrinsic (the invoice/card id) rather than to the array index.

JohnathanWhite (commented):

src/store/transforms/encrypt.ts
```diff
@@ -1,49 +1,237 @@
...
+  if (!hasProtectedValue(value)) {
     return value;
   }
-  try {
-    const encryptedText = value.replace(encryptedPrefix, '');
-    const result = Aes.decrypt(encryptedText, secretKey).toString(
-      CryptoJsCore.enc.Utf8,
+
+  if (typeof value !== 'string' || !isEncryptedValue(value)) {
+    throw new Error(`Expected encrypted protected value at ${context}`);
```
msalcala11, line 166: Older versions encrypted the whole APP/SHOP store rather than each field. Once we decrypt the store, the fields are plain strings, so this check throws and the app won't open. Previously, those fields loaded fine and were encrypted individually on the next save.

Can we keep that working for old saves without allowing plaintext in newer saves, and add an upgrade test?

msalcala11 (changes requested):

src/store/transforms/encrypt.ts
gabrielbazan7: Good catch. Thanks

PR #2279, opened 2026-08-10
Guidepoint: Fix - Strip internal build paths from release binaries

cmgustavo:
Jira: [RN-2909](https://bitpayprod.atlassian.net/browse/RN-2909)

Release iOS/Android binaries embedded absolute build-machine paths and developer usernames (e.g. `/Users/<dev>/dev/bitpay-app/...`) via `__FILE__/#file` macro strings and debug info in compiled native code. 
Stripping does not remove the macro strings, so add a Clang path-prefix-map that rewrites the repo root to "./" at compile time (computed per-build-machine; nothing absolute committed).

- **ios/Podfile**: append `-ffile-prefix-map` (C/C++/ObjC) and `-file-prefix-map` (Swift) to all Release pod targets in post_install.
- **android/build.gradle**: subprojects block appends `-ffile-prefix-map` to cppFlags/cFlags on each module's Release CMake build (state.executed guard avoids the already-evaluated react-native-gradle-plugin; Release-only).

**Verification:**

* **iOS**:  `grep -ra '/Users' Payload/BitPayApp.app | strings | egrep '^/Users/.+/.+' | awk -F/ '{print $3}' | sort -u`
* **Android**: `unzip -p base.apk | strings | grep '/Users/' | sort -u`


ios/Podfile
```diff
@@ -106,6 +106,33 @@ target 'BitPayApp' do
...
+      existing = config.build_settings[key]
+      current =
+        if existing.is_a?(Array)
+          existing
+        elsif existing.is_a?(String) && !existing.empty?
+          [existing]
+        else
+          ['$(inherited)']
+        end
+      config.build_settings[key] = (current + Array(flag_tokens)).join(' ')
```
JohnathanWhite: **`.join(' ')` breaks the prefix-map flag when the checkout path contains a space**

`repo_root` is interpolated into `clang_prefix_map`, then joined into a single space-separated build-setting string. For a checkout under e.g. `/Users/name/BitPay Wallet`, Xcode splits the setting at that space and the compiler receives a truncated `-ffile-prefix-map=/Users/name/BitPay` plus a stray `Wallet/=./` — malformed flags on every pod's Release configuration.

`OTHER_SWIFT_FLAGS` is affected the same way. Quoting and escaping the interpolated mapping, or keeping the setting as an array rather than joining, would fix it.

For contrast, the Android side is already safe: `android/build.gradle` uses `cppFlags.add(prefixMap)`, so the flag stays a single list element regardless of spaces.

JohnathanWhite (commented):

One finding.

ios/Podfile
cmgustavo: Fixed both halves: `append_build_flag` no longer joins into a string (the setting stays an array, one element per flag, matching the Android cppFlags.add behavior), and repo_root is now `.shellescaped` so a space in the checkout path is escaped for Xcode's whitespace split. OTHER_SWIFT_FLAGS gets the same treatment since it goes through the same helper.

JohnathanWhite (approved):

PR #2301, opened 2026-08-17
LogProvider: Fix - Defer LogProvider state updates to avoid setState-…

cmgustavo:
Jira: [RN-2888](https://bitpayprod.atlassian.net/browse/RN-2888)

…during-render warning

Fix:

```
"Cannot update a component (LogProvider) while rendering a different
   component (KeyOverview)."
```

Fix it at the source of the setState rather than at each render site: LogProvider now coalesces logManager notifications and applies them after the current render/commit via requestAnimationFrame, cancelling any pending frame on unmount. This covers every screen that logs during render and batches chatty log bursts into a single state update.

### DIFF
--- a/src/contexts/LogContext.tsx
+++ b/src/contexts/LogContext.tsx
@@ -3,6 +3,7 @@ import React, {
   useContext,
   useState,
   useEffect,
+  useRef,
   ReactNode,
 } from 'react';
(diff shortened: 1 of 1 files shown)


JohnathanWhite (approved):

JohnathanWhite:
https://bitpayprod.atlassian.net/browse/RN-2888

PR #2309, opened 2026-08-18
chore: drop 32-bit x86 ABI from Android builds

JohnathanWhite:
No consumer Android device ships 32-bit-only x86 (the Atom phone line ended in 2016) and Play has required 64-bit since Aug 2019, so the x86 slice is dead weight in every AAB. x86_64 stays for ChromeOS, Intel-Mac emulators, and Play pre-launch test devices.


gabrielbazan7 (approved):
LGTM

JohnathanWhite:
https://bitpayprod.atlassian.net/browse/RN-2893

PR #2310, opened 2026-08-18
Performance: chore - drop x86_64 from Android release builds

JohnathanWhite:
Follows dropping 32-bit x86. x86_64 is emulator-only, so every release APK carried a full extra copy of every native library that no real device executes. The debug build type adds it back via AGP's abiFilters union, so emulator development is unaffected.

Also documents why enableProguardInReleaseBuilds must stay false for now: proguard-rules.pro is empty and this app reflects heavily through bitcore/BWC, Braze, and Sumsub so enabling R8 needs its own change with a full QA pass.


gabrielbazan7:
follow-up: https://github.com/JohnathanWhite/bitpay-app/pull/13

gabrielbazan7:

PR #2311, opened 2026-08-18
Performance: feat - mount the DKLS TSS WebView on first use

JohnathanWhite:
The host was mounted unconditionally at app root, so every launch built a ~1.2MB boot HTML string from the vendored base64 blobs, shipped it over the bridge and spawned a WebView process for a feature most sessions never touch. It now renders null until the first callWorker call, which only happens during a TSS operation.


gabrielbazan7:
follow-up: [perf/dkls-lazy-mount-followup](https://github.com/gabrielbazan7/bitpay-app/tree/perf/dkls-lazy-mount-followup)

PR #2312, opened 2026-08-18
Performance: feat - parallelize launch and refresh network requests

JohnathanWhite:
Balances were awaited behind roughly nine serial round trips: one token-list GET per supported chain, two fiat-rate GETs, then per-key and per-read-only-wallet status calls. These are independent, so they now run together. Per-chain and per-chunk failures are isolated rather than aborting the whole effect, and rates this cycle did not produce are carried forward instead of erased.


gabrielbazan7:
follow-up: https://github.com/JohnathanWhite/bitpay-app/pull/11

gabrielbazan7:

PR #2313, opened 2026-08-18
Performance: feat - resolve BwcProvider lazily instead of at module scope

JohnathanWhite:
Root.tsx and the persist transforms both called BwcProvider.getInstance() at module scope, which forced the whole bitcore-wallet-client graph - four bitcore-lib variants plus crypto-wallet-core, whose elliptic-curve precompute tables run real work at module init - to load during bundle evaluation, on the critical path to first paint. Metro inlineRequires already defers this when the only references are inside functions.


gabrielbazan7:
The diagnosis is correct—loading BWC/bitcore takes ~300 ms—but this diff doesn’t defer it. Even after fixing all top-level imports, the cost merely shifts to Redux rehydration, with no measurable first-paint improvement. Recommendation: don’t merge.

PR #2314, opened 2026-08-18
Performance: fix - stop the persist transform mutating live wallet state

JohnathanWhite:
The bindWalletKeys inbound transform ran `delete wallet.transactionHistory` on wallet objects it does not own. redux-persist passes inbound transforms the live store slice by reference, so this mutated in-memory Redux state: the tx-history cache was wiped on every persist flush, moments after the reducer wrote it. The read-side cache check in GetTransactionHistory was therefore dead, and history was re-downloaded on every wallet or account view - one request per wallet on account views.

Restoring that cache re-arms a safety gate that had been dead as a side effect of the same bug: send.ts blocks an ETH send when `hasConfirmingTxs` is set, to avoid a nonce collision with a pending transaction. Nothing recomputes that flag unless a details screen is mounted or the user pulls to refresh, so a cached value can be arbitrarily old. Trusting a stale value locks the user out of sending; discarding it disables the protection for exactly the case it exists for. So the flag now carries a fetchedAt stamp and, when stale, is re-verified against the network before the send is allowed or blocked. On a fetch failure it fails closed.

For ERC20 sends the re-verification targets the linked native wallet, because the flag for a token wallet is derived from the linked wallet's history - nonces are account-level - and re-fetching the token wallet would recompute from the same stale cache. Load-more pages carry the previous flag and its age forward rather than reporting false, which would clobber a correct value after an ordinary scroll.

### DIFF
--- a/src/navigation/wallet/screens/AccountDetails.tsx
+++ b/src/navigation/wallet/screens/AccountDetails.tsx
@@ -435,6 +435,9 @@ const AccountDetails: React.FC<AccountDetailsScreenProps> = ({route}) => {
       transactions: any[];
       loadMore: boolean;
       hasConfirmingTxs: boolean;
+      // Carried through from GetTransactionHistory; send.ts uses it to decide
+      // whether hasConfirmingTxs still needs re-verification.
+      fetchedAt?: number;
     };
   }>({});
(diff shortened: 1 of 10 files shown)


gabrielbazan7:
follow up: https://github.com/JohnathanWhite/bitpay-app/pull/10

gabrielbazan7:

PR #2315, opened 2026-08-18
Performance: fix - throttle root persist writes and flush before suspend

JohnathanWhite:
redux-persist's throttle defaulted to 0, so the persist pipeline ran at maximum frequency; whenever the WALLET slice is dirty that pipeline re-runs pure-JS AES over six key fields per key and re-serializes merged BWC client objects. Throttling coalesces it. Because throttled writes can be starved indefinitely while several slices keep churning, persistence is now also flushed when the app backgrounds and after key create/import/delete. Also stops serializing the whole store twice at rehydrate for a log line.

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -296,18 +296,17 @@ export const reduxStorage: Storage = {
   },
 };
 
-// The live persistor, so the middleware below can force a write for actions whose
-// data must not sit in the throttled queue.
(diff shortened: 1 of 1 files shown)


gabrielbazan7:
follow up: https://github.com/JohnathanWhite/bitpay-app/pull/7

gabrielbazan7:

PR #2316, opened 2026-08-18
Performance: feat - skip no-op portfolio populate updates and narrow slice selectors

JohnathanWhite:
UPDATE_POPULATE_PROGRESS is dispatched from a 200ms poll and returned a brand-new slice object on every tick even when every field was identical, re-rendering every component that subscribed to the whole PORTFOLIO slice. The reducer now bails out when nothing changed, and the five whole-slice subscribers select only the fields they read - three of them just a boolean, so they re-render when populate starts and stops rather than on every tick.


msalcala11 (approved):

gabrielbazan7:

PR #2317, opened 2026-08-18
Performance: feat - token options memoization

JohnathanWhite:
### Issue

  Seven screens each built the merged token map inline **inside** `useAppSelector`:

  ```js
  useAppSelector(({WALLET}) => ({
    ...BitpaySupportedTokenOptsByAddress,
    ...tokenOptionsByAddress,
    ...WALLET.customTokenOptionsByAddress,
  }))
  ```

  `useSelector` runs on every store dispatch, so this spread a multi-thousand-key object on every dispatch, for every mounted instance. And because the result was always a fresh reference, the equality check could never bail — so the component
  re-rendered on every dispatch too. Worst on the 1,500–1,900 line KeyOverview/AccountDetails screens and on per-row components (`ContactIcon`, `MultipleOutputsTx`).

  A shared `useTokenOptionsByAddress` hook now selects only the custom-token map (a stable reference) and merges in `useMemo`, so the identity stays stable until the inputs actually change.

  **Also fixes a bug:** `ContactIcon` spread the const it was declaring instead of the context value, so the 1inch token map was silently dropped there. eslint had been flagging the unused binding.

  **Second commit** applies the hook to `GlobalSelect` and cleans up its render path: it walked the entire token registry and ran ~8 sequential wallet filter passes bare in the render body, so all of it re-ran on every search keystroke and every
  modal toggle. That's now memoized, it selects one WALLET field instead of the whole slice, and its two FlashLists key on stable ids instead of array indices (index keys over filtered, changing data break recycling and can reuse a row for the
  wrong wallet).

  ### How to test

  1. Token icons and names resolve correctly on KeyOverview, AccountDetails, KeySettings, Add Custom Token, transaction detail with multiple outputs, and the contacts list. Some icons may now appear where a generic fallback showed before —
  that's the ContactIcon fix, not a regression.
  2. Open the wallet/coin picker: type quickly in search, toggle chain filters, scroll. Rows must show the **correct** wallet — a wrong-row or duplicated row would indicate a list-key problem.
  3. Add a custom token and confirm it still appears everywhere it should.
  4. Send and swap flows still resolve token metadata correctly.

### DIFF
--- a/src/components/list/ContactRow.tsx
+++ b/src/components/list/ContactRow.tsx
@@ -52,7 +52,7 @@ interface Props {
 const ContactRow = ({contact, onPress}: Props) => {
   const theme = useTheme();
   const underlayColor = theme.dark ? '#121212' : GhostWhite;
-  const {coin: _coin, name, email, address, chain, tokenAddress} = contact;
+  const {coin: _coin, name, email, address, chain} = contact;
   const coin = getCurrencyAbbreviation(_coin, chain);
   return (
     <ContactContainer underlayColor={underlayColor} onPress={onPress}>
@@ -64,7 +64,6 @@ const ContactRow = ({contact, onPress}: Props) => {
             size={45}
             chain={chain}
             address={address}
-            tokenAddress={tokenAddress}
           />
         </ContactImageContainer>
         <ContactColumn>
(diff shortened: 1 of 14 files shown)


gabrielbazan7:
Follow-up to 1903baa3 and 40ee6527: https://github.com/JohnathanWhite/bitpay-app/pull/6

gabrielbazan7:

PR #2356, opened 2026-08-31
Backup: Fix - Make redux-persist file writes atomic and idempotent on iOS

cmgustavo:
Fix:

**`"persist-root.json.tmp" couldn't be moved ... an item with the same name already exists`** When the `final → .bak` rotation failed it was logged and swallowed, leaving `persist-root.json` in place, so the following `moveFile(tmp → final)` hit an occupied destination. iOS `NSFileManager.moveItem` throws instead of overwriting. Both moves now go through `moveOverwriting()`, which unlinks the destination first.

**`"persist-root.json.tmp" couldn't be moved ... the former doesn't exist, or the folder containing the latter doesn't exist`**
`RNFS.writeFile` rejects with `ENOENT` when the folder is missing — a different message — so the directory existed at write time. The temp file and its folder disappear *after* the write: iOS can purge `CachesDirectoryPath` on any suspended `await`. The rotation now happens before the temp write (3 awaits between write and move → 1) and the write+move pair retries once, recreating the directory.

### DIFF
--- a/src/store/backup/fs-backup.spec.ts
+++ b/src/store/backup/fs-backup.spec.ts
@@ -35,6 +35,74 @@ jest.mock('../../store/log/initLogs', () => ({
 const mockedRNFS = RNFS as jest.Mocked<typeof RNFS>;
 
 // ─────────────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


cmgustavo:
Already fixed in 14.43

PR #2357, opened 2026-08-31
Buy/Sell: Fix - error handling when no available offers

Gamboster:
Fixes a crash in the buy offer-selection flow and aligns the sell flow's "no offers" warning logic with it.

Buy flow: Array.prototype.reduce() was called without an initial value on filteredOffers even when allExchangesFinished was false and the array was empty, causing `TypeError: Array.prototype.reduce() requires an initial value with empty array` (reported in Sentry)

Sell flow: added the equivalent allSellExchangesFinished/hasParticipatingExchange check (mirroring the buy flow, scoped to Moonpay/Ramp/Simplex) so the "no offers" warning isn't shown prematurely while one of the sell exchanges is still loading.

### DIFF
--- a/src/navigation/services/components/externalServicesOfferSelector.tsx
+++ b/src/navigation/services/components/externalServicesOfferSelector.tsx
@@ -2299,6 +2299,7 @@ const ExternalServicesOfferSelector: React.FC<
     sellOffers.moonpay.amountReceivingAltFiatCurrency = undefined;
     sellOffers.moonpay.fiatMoney = undefined;
     sellOffers.moonpay.expanded = false;
+    setFinishedMoonpay(true);
     setUpdateViewSell(Math.random());
   };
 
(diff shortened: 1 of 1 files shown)


src/navigation/services/components/externalServicesOfferSelector.tsx
```diff
@@ -3077,6 +3082,21 @@ const ExternalServicesOfferSelector: React.FC<
...
+      const allSellExchangesFinished =
+        moonpayReady && rampReady && simplexReady;
+
+      const hasParticipatingExchange =
+        finishedMoonpay !== null ||
+        finishedRamp !== null ||
+        finishedSimplex !== null;
+
+      if (!hasParticipatingExchange) {
+        return;
```
JohnathanWhite, line 3097: **All-null exchange state leaves the selector loading forever**

`hasParticipatingExchange` is false only when all three `finished*` flags are `null`. This early `return` then skips both `setSelectedOfferLoading(false)` and the "no offers" warning, so the selector spins indefinitely.

Worth noting this is a behaviour change rather than a pre-existing bug: with all three flags `null`, `moonpayReady`/`rampReady`/`simplexReady` are each true, so `allSellExchangesFinished` is true and the pre-change code *would* have shown the warning and cleared loading.

Reachable for an unsupported currency or region where no exchange participates. Treating all-null as a completed no-offer result — or clearing loading before returning — would restore the old behaviour.

JohnathanWhite (commented):

One finding.

JohnathanWhite (approved):

PR #2367, opened 2026-09-02
Logs: Fix - record post-startup logs instead of dropping them

cmgustavo:
Logs emitted after store initialization were silently discarded: `initLogs.add` only buffered them, and that buffer is drained once at store creation and never again — so every log from `fs-backup.ts` and `transforms.ts` went nowhere. Backup failures like `Backup write failed - ENOENT` never reached the `LOG` slice or the `persist:logs` MMKV key, so they showed in neither section of Session Log and were missing from the log file users email to support; only the Sentry breadcrumb survived.

The fix appends to `persist:logs` directly rather than dispatching, because `ADD_PERSISTED_LOG` mutates `LOG` state and triggers a `persist:root` write — itself a path that logs here, so dispatching would close the loop. `addLog` in `store/index.ts` had that exact loop with MMKV failures and now routes through the same path.

### Steps to reproduce (only Android):

1. Create and remove cache folders: 

```
adb shell run-as com.bitpay.wallet mkdir -p cache/bitpay; adb shell run-as com.bitpay.wallet rm -rf cache/bitpay/redux; adb shell run-as com.bitpay.wallet touch cache/bitpay/redux; adb shell run-as com.bitpay.wallet ls -l cache/bitpay
```

2. Wallet Settings → Hide Wallet: tap 3-4 times

```
[Error] Backup write failed - ENOENT: open failed: ENOTDIR (Not a directory), open '/data/user/0/com.bitpay.wallet/cache/bitpay/redux/persist-root.json.tmp'
```

3. Finish testing: 

```
adb shell run-as com.bitpay.wallet rm -f cache/bitpay/redux
```

### Side Notes

`persist:logs` is only pruned at startup (`CLEAR_LOGS`, 7-day cutoff) and every append re-serializes the whole array, so a session stuck in a failure loop grew it unbounded — now that backup errors actually reach it, that's a real path. Capped at the newest 500 entries on write.

### DIFF
--- a/src/store/log/initLogs.ts
+++ b/src/store/log/initLogs.ts
@@ -1,13 +1,50 @@
+import {LogEntry, sanitizeLogMessage} from './log.models';
 import type {AddLog} from './log.types';
+import {logManager} from '../../managers/LogManager';
+import {storage} from '../index';
 
 // For storing logs before the store is initialized
 const initLogs: AddLog[] = [];
+let drained = false;
+
+// CLEAR_LOGS only prunes persist:logs at startup, so a session stuck in a
+// failure loop would grow the array all session long — and every append
+// re-serializes the whole thing
+const MAX_PERSISTED_LOGS = 500;
+
+// Single writer for persist:logs, shared with the ADD_PERSISTED_LOG reducer so
+// the cap and the redaction can't diverge between the two paths
+export const appendPersistedLog = (entry: LogEntry): LogEntry => {
+  const sanitized = {...entry, message: sanitizeLogMessage(entry.message)};
+
+  // Session Log reads logManager, not LOG.logs — without this, persisted logs
+  // are missing from both on-screen sections and from the exported current
+  // session. Not a second dispatch: ADD_LOG dirties LOG state, which makes
+  // redux-persist write persist:root, itself a path that logs here
+  logManager.addLog(sanitized);
+
+  try {
+    const persistedLogs = storage.getString('persist:logs') || '[]';
+    storage.set(
+      'persist:logs',
+      JSON.stringify(
+        [...JSON.parse(persistedLogs), sanitized].slice(-MAX_PERSISTED_LOGS),
+      ),
+    );
+  } catch {}
+  return sanitized;
+};
 
 export const add = (log: AddLog) => {
+  if (drained) {
+    appendPersistedLog(log.payload);
+    return;
+  }
   initLogs.push(log);
 };
 
 export const drainAndDispatch = (dispatch: (action: AddLog) => void) => {
+  drained = true;
   if (initLogs.length === 0) {
     return;
   }
--- a/src/navigation/tabs/settings/about/screens/SessionLog.tsx
+++ b/src/navigation/tabs/settings/about/screens/SessionLog.tsx
@@ -168,10 +168,13 @@ const SessionLogs = ({}: SessionLogsScreenProps) => {
   const currentSessionStartTime = logs.length
     ? new Date(logs[0].timestamp)
     : new Date();
-  const [filteredPersistedLogs, setFilteredPersistedLogs] = useState(
-    [] as LogEntry[],
-  );
+  // persist:logs also holds this session's entries (fs-backup/transforms write
+  // there directly), so split by timestamp — not by source — to keep them out
+  // of the "Previous Sessions" section and out of the export's previous block
   const [persistedLogs, setPersistedLogs] = useState([] as LogEntry[]);
+  const filteredPersistedLogs = persistedLogs.filter(
+    log => log.level <= filterLevel,
+  );
 
   const printLogs = (logsToPrint: LogEntry[]) =>
     logsToPrint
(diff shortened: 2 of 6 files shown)


src/store/log/initLogs.ts
```diff
@@ -1,13 +1,40 @@
...
+      JSON.stringify(
+        [...JSON.parse(persistedLogs), log.payload].slice(-MAX_PERSISTED_LOGS),
+      ),
+    );
+  } catch {}
+};
 
 export const add = (log: AddLog) => {
+  if (drained) {
+    appendPersistedLog(log);
```
JohnathanWhite: **Post-startup logs are written to MMKV but never appear in Session Log for the rest of the session**

After `drainAndDispatch` runs, `add()` writes straight to `persist:logs` and returns without dispatching, so the entry never reaches Redux `LOG.logs`. `SessionLog.tsx` builds its two on-screen lists from (a) Redux `logs` for "Current Session" and (b) `persist:logs` filtered by `new Date(log.timestamp) < currentSessionStartTime` for "Previous Sessions" (`SessionLog.tsx:284-291`). A log written after startup has a current-session timestamp, so it is filtered out of "Previous Sessions" and is absent from Redux — it is invisible in the UI until the next app launch. Worse, the email/share export uses the *unfiltered* `persistedLogs`, so these entries do get exported but are printed under the `Previous Sessions` header even though they belong to the current session. Since these are precisely the MMKV/backup failure logs support needs, consider keeping a non-persisting dispatch (e.g. `ADD_LOG`, which does not touch storage) alongside the MMKV append, or have `SessionLog` partition by timestamp rather than by source.

src/store/log/initLogs.ts
```diff
@@ -1,13 +1,40 @@
...
+import {storage} from '../index';
 
 // For storing logs before the store is initialized
 const initLogs: AddLog[] = [];
+let drained = false;
+
+// CLEAR_LOGS only prunes persist:logs at startup, so a session stuck in a
+// failure loop would grow the array all session long — and every append
+// re-serializes the whole thing
+const MAX_PERSISTED_LOGS = 500;
```
JohnathanWhite, line 13: **The 500-entry cap is enforced on only one of the two writers to persist:logs, so one append can delete older logs**

`appendPersistedLog` applies `.slice(-MAX_PERSISTED_LOGS)`, but the other writer to the same key — `ADD_PERSISTED_LOG` in `log.reducer.ts:43-47` — appends with no cap at all. The effective size of `persist:logs` therefore depends on which path wrote last: once the reducer has grown the array past 500 (it is only ever pruned by `CLEAR_LOGS`, which prunes by age at startup, not by count), the very first post-startup `add()` call silently truncates it to the newest 500, discarding previous-session entries that were being retained on purpose. Either apply the cap in both writers or move the append behind a single shared helper so the two paths cannot disagree.

JohnathanWhite (commented):

Three findings. The overall direction of the fix looks right; these are about the edges.

src/store/log/initLogs.ts
```diff
@@ -1,13 +1,40 @@
...
+// Appends to MMKV instead of dispatching: ADD_PERSISTED_LOG mutates LOG state,
+// which makes redux-persist write persist:root, which is itself a code path
+// that logs here — dispatching would close that loop
+const appendPersistedLog = (log: AddLog) => {
+  try {
+    const persistedLogs = storage.getString('persist:logs') || '[]';
+    storage.set(
+      'persist:logs',
+      JSON.stringify(
+        [...JSON.parse(persistedLogs), log.payload].slice(-MAX_PERSISTED_LOGS),
```
JohnathanWhite: **This write path bypasses sanitizeLogMessage, which the reducer applies before persisting**

`ADD_PERSISTED_LOG` runs `sanitizeLogMessage(action.payload.message)` before writing to `persist:logs` (`log.reducer.ts:34-47`); `appendPersistedLog` writes `log.payload` verbatim. Post-startup logs are therefore persisted unsanitised. In practice nothing changes *today* because `sanitizeLogMessage` is already a no-op — it calls `message.replace('/xpriv.*/', '[...]')` with a **string**, not a `RegExp`, so it only matches the literal text `/xpriv.*/` — but that is a latent bug elsewhere, and the moment it is fixed this path will still be unsanitised. Route the payload through the same sanitiser (or fix and share it) so the two writers cannot diverge on redaction.

src/store/log/initLogs.ts
cmgustavo: Good catch, fixed — but the premise is one layer off, which changed the fix.

Session Log (and the emailed file) reads the `logManager` singleton, not the Redux `LOG.logs` slice — that slice is only used by the internal Debug screen. So an extra `ADD_LOG` dispatch would not have made these visible, and it would have reopened the write loop this PR set out to avoid. Post-startup logs now go to `logManager` instead, which is memory-only: they show up under "Current Session" right away, no storage write involved.

Also took your second suggestion: Session Log now splits the persisted logs by timestamp rather than by source, so nothing from this session ends up printed under the "Previous Sessions" header in the export.

Side benefit: this also makes visible a batch of logs that were already being dropped from the UI before this PR (the ones dispatched from contacts, wallet import and Root).

src/store/log/initLogs.ts
cmgustavo: Right, fixed. There is now a single function that writes to `persist:logs`, and the reducer calls it too, so both paths share the same cap. Covered by a test.

src/store/log/initLogs.ts
cmgustavo: Fixed — same shared function, so both paths sanitize identically now.

I left the broken regex itself alone: as you noted it is a no-op today, and making it actually match would swallow the rest of any log line that merely mentions `xpriv`. That deserves its own change rather than a guess buried in this PR, so I will track it separately.

JohnathanWhite (approved):

PR #2369, opened 2026-09-03
Notifications: Fix - open tx from tapped notification

gabrielbazan7:
Jira: [RN-2723](https://bitpayprod.atlassian.net/browse/RN-2723)

### DIFF
--- a/src/store/app/app.effects.ts
+++ b/src/store/app/app.effects.ts
@@ -1285,6 +1285,45 @@ export const incomingLink =
         handler = () => {
           navigationRef.navigate(WalletScreens.CREATION_OPTIONS, params);
         };
+      } else if (params.walletId) {
+        handler = async () => {
+          const {
+            WALLET: {keys},
+          } = getState();
+          const {wallet, keyId} = await findWalletByIdHashed(
+            keys,
+            params.walletId,
+            null,
+          );
+
+          if (!wallet || !keyId) {
+            return;
+          }
+
+          const key = keys[keyId];
+          const tokenAddress =
+            params.tokenAddress && params.tokenAddress !== 'null'
+              ? params.tokenAddress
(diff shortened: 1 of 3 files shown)


cmgustavo (approved):
ACK!

JohnathanWhite (commented):

One finding.

src/store/app/app.effects.ts
```diff
@@ -1285,6 +1285,45 @@ export const incomingLink =
...
+            params.txid && params.txid !== 'null' ? params.txid : undefined;
+          const tokenWalletId =
+            `${wallet.credentials.walletId}-${tokenAddress}`.toLowerCase();
+          const targetWallet =
+            (tokenAddress &&
+              key.wallets.find(
+                (w: Wallet) =>
+                  w.credentials.walletId.toLowerCase() === tokenWalletId,
+              )) ||
+            wallet;
```
JohnathanWhite, line 1318: **`|| wallet` fallback sends a token notification to the base wallet**

When `params.tokenAddress` is set but nothing in `key.wallets` matches `tokenWalletId`, this falls back to the base `wallet`. Navigation then carries the token transfer's `txid` into the base wallet's detail screen, where the lookup can't find it — the user taps a token notification and lands on the wrong wallet showing no transaction.

Reachable when the token wallet hasn't loaded yet, was removed, or `tokenAddress` is stale. Consider returning without navigating (or deferring until the token wallet resolves) when `tokenAddress` was supplied but unresolved, rather than silently degrading to the base wallet.

JohnathanWhite (approved):

src/store/app/app.effects.ts
gabrielbazan7: https://github.com/bitpay/bitpay-app/pull/2382

PR #2370, opened 2026-09-03
TSS: Fix - align with upstream BWS/BWC TSS changes 

gabrielbazan7:


src/store/wallet/effects/tss-send/tss-send.ts
```diff
@@ -223,8 +218,10 @@ const signInput = async (params: {
           `[TSS Sign] Restoring saved session — sessionId: ${sessionId}, lastRound: ${savedSession.round}`,
         );
         try {
-          await tssSign.restoreSession({session: savedSession.exported});
-          tssSign.id = sessionId;
+          await tssSign.restoreSession({
+            session: savedSession.exported,
```
JohnathanWhite, line 222: **Draft blocker: this depends on an unreleased BWC — the pinned 11.10.6 restoreSession still splits on ':' and takes no password**

This PR deletes the `extractSignData`/`buildSessionForRestore` workaround **and** the `tssSign.id = sessionId` override, and starts passing `password` to `restoreSession` — but `package.json` still pins `@bitpay-labs/bitcore-wallet-client` at `11.10.6`, unchanged from `develop`. In that version `restoreSession` is `restoreSession(params: {session: string})` and its body is `const [id, sigSession] = session.split(':'); this.id = id;` (`src/lib/tsssign.ts:160-175`) — there is no `password` parameter, so the one passed here is silently dropped. Meanwhile `generateSessionId` returns `` `${txp.id}:input${i}` `` (line 170 of this file), so `exportSession()` produces `"<txpId>:input0:<base64>"`; the destructured `split(':')` therefore yields `id = "<txpId>"` and `sigSession = "input0"`, discarding the sign data entirely and leaving `this.id` truncated so `subscribe()` and the round POSTs target `/v1/tss/sign/<txpId>` instead of `/v1/tss/sign/<txpId>:input0`. That is exactly the bug the deleted comments described. On the pinned dependency every session restore fails: the primary path falls through to `clearSigningSession` + `start()`, BWS answers `TSS_ROUND_ALREADY_DONE`/`TSS_ROUND_MESSAGE_EXISTS`, and the user is told to delete the proposal — so an interrupted TSS send becomes unrecoverable. (Supporting signal: neither `TSS_SESSION_EXPIRED` nor `TSS_MISMATCH_VERSION` appears anywhere in BWC 11.10.6.) Please bump BWC to the version that carries the upstream `restoreSession` fix in this same PR, and state the minimum version in the description.

JohnathanWhite (commented):

PR #2377, opened 2026-09-07
Translations: Feat - full coverage in eight languages, plus tooling

gabrielbazan7:
**Dependency upgrade:** `i18next` 21.6.4 → 25.8.14 and `react-i18next` 12.2.0 → 16.5.4.
 **Crowdin is gone:** translations live in the repo and `yarn translation:check` runs in
  CI, erroring on any missing translation. From here on, a PR adding a user-facing string
  must ship all 8 translations or CI blocks it. Workflow in `locales/README.md`.


package.json
```diff
@@ -103,8 +102,9 @@
     "fast-text-encoding": "1.0.6",
     "flatted": "3.2.5",
     "https-browserify": "1.0.0",
-    "i18next": "21.6.4",
+    "i18next": "25.8.14",
```
JohnathanWhite, line 105: **A four-major i18next and react-i18next upgrade is riding inside a translations PR**

This bumps `i18next` 21.6.4 → **25.8.14** and `react-i18next` 12.2.0 → **16.5.4** — four majors each — and correspondingly drops `compatibilityJSON: 'v3'` from `i18n.js` (v3 JSON was removed from i18next in v23, so the removal is mandatory rather than optional). That touches every `t()`, `useTranslation` and `Trans` call site in the app plus a regenerated `yarn.lock`, and it is not something the title or description prepares a reviewer for. Please call it out explicitly so it gets deliberate sign-off and a device smoke test rather than being read as catalog churn. Two related changes in this diff are genuine fixes worth noting in the description as well: adding `keySeparator: false` / `nsSeparator: false` means keys containing `.` or `:` (there are many, since keys are literal English phrases) now resolve instead of being silently split into a namespace and a key path, and the `intl-pluralrules` import supplies the `Intl.PluralRules` implementation the new v4 plural suffixes need on Hermes.

src/components/modal/import-ledger-wallet/import-account/SelectWalletsToImport.tsx
```diff
@@ -185,21 +178,18 @@ export const SelectWalletsToImport: React.FC<Props> = props => {
...
-              Connect {selectedWallets} Wallet
-            </Button>
-          )}
+          <Button
+            testID={
+              selectedWallets > 1
+                ? 'ledger-connect-wallets-button'
+                : 'ledger-connect-wallet-button'
+            }
+            accessibilityLabel={`Connect ${selectedWallets} ${
```
JohnathanWhite: **accessibilityLabel left as a hardcoded English template while the visible label is translated**

The button caption became `t('Connect {{count}} wallet', {count: selectedWallets})`, but `accessibilityLabel` is still an untranslated template literal that hand-rolls its own English pluralisation. Screen-reader users in the eight locales this PR adds will hear English while sighted users see their own language, and the two strings can now disagree. Reuse the translated value — e.g. `accessibilityLabel={t('Connect {{count}} wallet', {count: selectedWallets})}` — and drop the manual `wallets`/`wallet` branch, since i18next is already selecting the right plural form.

src/navigation/tabs/home/components/Crypto.tsx
```diff
@@ -626,11 +623,23 @@ const Crypto = () => {
...
+            height={carouselHeight}
             autoPlay={false}
             data={cardsList.list}
             scrollAnimationDuration={0}
-            renderItem={_renderItem}
+            renderItem={({item}) => (
+              <CarouselItemContainer
+                onLayout={({nativeEvent}) =>
+                  setCarouselHeight(current =>
+                    Math.max(
```
JohnathanWhite: **Carousel height only ever grows, so the tallest card ever measured pins the carousel for the rest of the session**

`setCarouselHeight(current => Math.max(current, …))` makes the height monotonically non-decreasing. The ratchet is what stops an onLayout/setState feedback loop, so it is the right shape — but it also means the height never comes back down after the tallest card unmounts, after a rotation, or after the user reduces their text-size setting, leaving permanent dead space below the carousel. Deriving the height from the current `cardsList.list` measurements (e.g. a per-item height map, reset when the list identity changes) would keep the convergence property without the one-way ratchet. Separately, `renderItem` is now an inline closure rebuilt on every render, so it is a new prop identity each pass and defeats any memoisation inside the carousel — worth wrapping in `useCallback`.

package.json
```diff
@@ -43,9 +43,8 @@
     "reset-allowed-url-prefixes": "node ./scripts/allowed-url-prefixes.js reset",
     "reset-braze": "node ./scripts/braze-config-reset.js",
     "reset-google-maps": "node ./scripts/google-maps-reset.js",
-    "translation:extract": "yarn run i18next 'src/**/*.{js,jsx,ts,tsx}'",
-    "translation:upload": "crowdin upload sources --identity crowdin-credentials.yml",
-    "translation:download": "crowdin download --identity crowdin-credentials.yml"
+    "translation:extract": "npx i18next-cli extract",
+    "translation:check": "node --test scripts/check-translations.test.js && node scripts/check-translations.js"
```
JohnathanWhite, line 47: **Crowdin is removed and CI now hard-fails on any missing translation, with no TMS left to produce one**

`crowdin.yml` is deleted and the `translation:upload` / `translation:download` scripts are gone, while `translation:check` is wired into `.github/workflows/test.yml` and `locales/README.md` states that missing translations fail CI. The consequence for everyday work is worth stating explicitly in the description: from here on, any PR that adds a single user-facing string must also hand-produce all eight translations or CI blocks it — there is no longer a translation pipeline to fall back on. That may well be the intent (the README documents an AI-assisted workflow), but it is a process change that the whole team inherits, so it should be an explicit decision rather than a side effect. Verified locally that the checker itself is sound: `node scripts/check-translations.js` reports 0 errors against the nine catalogs at this head under the pinned i18next 25.8.14, and `node --test scripts/check-translations.test.js` passes 7/7.

JohnathanWhite (commented):

Four findings. The locale catalogues themselves checked out clean — details in the summary below.

---

**Locale consistency check (separate from the Codex pass): clean.** All 9 catalogues parse, 0 duplicate keys, 0 placeholder mismatches (`{{…}}`, `%s`, `%1$s`) across all 8 translations vs English. Key-set asymmetries are correct CLDR plural categories across 11 plural families (ja/zh `_other`; es/fr/pt `_one/_many/_other`; ru `_one/_few/_many/_other`). Every literal `t()` key across 977 source files resolves in `en.json`. Ran the PR's own new checker under the pinned i18next 25.8.14: 0 errors, and its tests pass 7/7.

package.json
gabrielbazan7: Correct, and it should have been in the description.

package.json
gabrielbazan7: Yes, that's correct. I've just added it to the description.

JohnathanWhite (commented):
**Verdict: shippable as is.**

The i18next 21 -> 25 / react-i18next 12 -> 16 jump is the risky part of this and it looks correctly handled:

- catalogs use v4 plural suffixes, consistent with v22 moving to `Intl.PluralRules`
- `intl-pluralrules` is required before i18next init
- interpolation escaping stays disabled for React
- `keySeparator: false` / `nsSeparator: false` protect the literal-sentence keys — which matters a lot here, since keys contain `.` and `:`
- Suspense in react-i18next v16 is a non-issue because resources are bundled synchronously

**One follow-up, explicitly not blocking: the new CI gate can't see missing keys.**

`scripts/check-translations.js` reads only `locales/<lang>/translation.json` (line 186) and derives everything from `Object.keys(en)` (line 51) — it never scans `src/`. So it enforces catalog↔catalog consistency, and does that thoroughly: interpolation mismatch, markup balance, brace pairing, plural suffixes, whitespace. But a `t('...')` whose key is absent from `en` is invisible to it by construction, and renders as the raw key.

Six live examples today, missing from all 9 catalogs:

- `1W Gain/Loss`, `1M Gain/Loss`, `3M Gain/Loss`, `1Y Gain/Loss`, `5Y Gain/Loss` — `src/navigation/tabs/home/components/AssetsGainLossDropdown.tsx`
- `Ready to Continue` — `src/navigation/wallet/screens/InviteCosigners.tsx`

All six were present in `locales/en` on develop and are dropped here, so rendered output is unchanged — English in every locale before and after. **No regression**, which is why this isn't blocking. But they are user-facing (Home portfolio period selector, cosigner invite screen) and they're precisely the class of gap this PR sets out to close. Worth either restoring the keys or adding a source-scan pass to the checker — `i18next-cli extract` already knows how to find them.

Nice work on the checker itself, incidentally — the interpolation and markup comparisons are more rigorous than the Crowdin setup they replace.

PR #2399, opened 2026-09-17
TSS: Fix - encrypt keychain in persisted state

gabrielbazan7:


gabrielbazan7:
added here: https://github.com/bitpay/bitpay-app/pull/2404

PR #2400, opened 2026-09-17
Wallet: Ref - stop persisting bwc client config

gabrielbazan7:
The wallet object in the store is a merge of the BWC client instance, so every persist was also writing the client's transport config (request, bulkClient, timeout, logLevel, bp_partner, _events). None of it survives a restart — bootstrapWallets builds a fresh client on every rehydrate — so it was ~60KB written on each save just to be discarded. Measured on a device with 47 wallets: persist:root went from 167KB to 102KB (-39%).


gabrielbazan7:
added here: https://github.com/bitpay/bitpay-app/pull/2404

PR #2404, opened 2026-09-18
Wallet: Fix - BIP-05

gabrielbazan7:
## What this does

Closes BIP-05 by encrypting the whole WALLET store instead of a hand-picked list of fields.
To make it affordable, it first stops persisting two things that never survive a restart anyway: the BWC client's transport config (`request`, `bulkClient`, `timeout`, `_events`…, re-created by `getClient()` on every rehydrate) and the fields `buildWalletObj` rebuilds (`balance`, `pendingTxps`, `isScanning`, `chainName`, `badgeImg`).

## Numbers

| approach | persist:root | per write |
|---|---|---|
| **today** — stringify + field-level on 3 fields per key | 167 KB | **~24 ms** |
| field-by-field, everything BIP-05 asks for | 167 KB | ~265 ms (210 AES calls) |
| whole store, no cleanup | 167 KB | **195 ms** (measured) |
| whole store after cleanup — **this PR** | ~84 KB | **~90 ms** |

So this costs ~66 ms more per write than today, at 4–11 writes per session, and covers every secret in the store instead of a list that drifts.

Encrypting field-by-field — what the ticket literally asks for — is the most expensive option of all, three times this PR: crypto-js derives a key on every call, so 210 small encryptions cost more than one pass over the whole store.


gabrielbazan7:
added here: https://github.com/bitpay/bitpay-app/pull/2405

